PR #13, opened 2016-06-07
Upgrade to master GeoNode

sarasafavi:
Updates Exchange to use the latest GeoNode master, which means a bump to Django 1.8 (among other things). `django-tastypie` has also been unpinned from our own deps to let us pull the appropriate version (matched to Django version) via GeoNode.

Also updates Exchange's use of `syncdb` to use `migrate` instead, as `syncdb` has been deprecated and is going away completely in Django 1.9.

Fair warning, I'm unfamiliar with everything else that changed with the move from our previous, much-older pinned GeoNode version to GeoNode master - so I'm not sure if there's a more complete QA process to be run here other than "works on my box".

**storytime:**

There is an issue with interdependent migrations. (related: [this](https://lists.osgeo.org/pipermail/geonode-devel/2016-February/000780.html) ML post). After the initial `makemigrations`, running `migrate` will fail with `ProgrammingError: relation "people_profile" does not exist`. Attempting to migrate `people` throws a different "relation does not exist" error (this time, for `sites`). I could list all the permutations of trying to migrate a particular app containing a missing relation first, but TL;DR: they all depend on something else, that depends on something else... The smallest & best fix I could find for this is to add the initial migration for `sites` as a dependency on `accounts`: i.e., add `('sites', '0001_initial'),` [here](https://github.com/GeoNode/geonode-user-accounts/blob/master/account/migrations/0001_initial.py#L34)).

~~I haven't PR'd that fix yet, pending further discussion. For now attempting `vagrant provision` with the bootstrap script in this PR will error on a "relation does not exist". Until the dependency on `sites` is added in to the initial migration in `geonode-user-accounts`, you can modify the migration file locally and manually apply migrations.~~ **Update**: this fix is now in geonode core, so `vagrant provision` should work fine with this PR's bootstrap.

### DIFF
--- a/cf/init_db.sh
+++ b/cf/init_db.sh
@@ -7,7 +7,7 @@ fi
 PASS = $GEONODE_ADMIN_PASSWORD
 
 echo "-----> Create database tables"
-python manage.py syncdb --noinput
+python manage.py migrate --noinput
 
 echo "-----> Create default admin user"
 echo "from geonode.people.models import Profile; Profile.objects.create_superuser('admin', 'admin@boundlessgeo.com', '$PASS')" | python manage.py shell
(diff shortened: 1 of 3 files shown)


sarasafavi:
PR for related migration fix: https://github.com/GeoNode/geonode-user-accounts/pull/4

PR #18, opened 2016-07-05
NODE-163: integrate osgeo importer

sarasafavi:
Ref: https://issues.boundlessgeo.com:8443/browse/NODE-163 (and related sub-tasks)

~~**CAVEAT: pending [NODE-198](https://issues.boundlessgeo.com:8443/browse/NODE-198), this is vector import only.** 
NODE-198 will fix geoTIFF imports in osgeo-importer~~

*local dev notes:*

* This is dependent on current open PRs on `django-osgeo-importer` upstream; this branch contains all PR commits if needed for immediate local testing: 
https://github.com/boundlessgeo/django-osgeo-importer/tree/exchange-integration

* Using osgeo-importer requires a celery worker to be active (even in dev) - to kick off a worker you can use the new convenience command, `django-celeryworker` (to be run concurrently with `django-runserver`)

*Screenshots:*

image
*Using django-importer UI to add new data*

image
*New: list of staged layers*

image
*Using osgeo-importer's UI*

image
*Link to access staged layer list, new in user menu*

### DIFF
--- a/exchange/templates/base.html
+++ b/exchange/templates/base.html
@@ -206,6 +206,14 @@
           </div>
           <div class="modal-body">
             <ul class="list-unstyled">
+              {% if REGISTRY %}
+              <li><a href="/registry/search/csw?service=CSW&version=2.0.2&request=GetCapabilities"><i class="fa fa-globe"></i> {% trans "Registry CSW Endpoint" %}</a></li>
+              {% endif %}
+              <li><a href="/catalogue/search/csw?service=CSW&version=2.0.2&request=GetCapabilities"><i class="fa fa-globe"></i> {% trans "Exchange CSW Endpoint" %}</a></li>
+              {% if USE_OSGEO_IMPORTER %}
+                  <li><a href="{% url "uploads-list" %}"><i class="fa fa-th-list"></i> {% trans "Staged Layers" %}</a></li>
+              {% endif %}
+              <li class="modal-divider"></li>
               <li><a href="{{ user.get_absolute_url }}"><i class="fa fa-user"></i> {% trans "Profile" %}</a></li>
               <li><a href="{% url "recent-activity" %}"><i class="fa fa-fire"></i> {% trans "Recent Activity" %}</a></li>
               <li><a href="{% url "messages_inbox" %}"><i class="fa fa-inbox"></i> {% trans "Inbox" %}</a></li>
--- a/exchange/templates/base.html
+++ b/exchange/templates/base.html
@@ -206,10 +206,6 @@
           </div>
           <div class="modal-body">
             <ul class="list-unstyled">
-              {% if REGISTRY %}
-              <li><a href="/registry/search/csw?service=CSW&version=2.0.2&request=GetCapabilities"><i class="fa fa-globe"></i> {% trans "Registry CSW Endpoint" %}</a></li>
-              {% endif %}
-              <li><a href="/catalogue/search/csw?service=CSW&version=2.0.2&request=GetCapabilities"><i class="fa fa-globe"></i> {% trans "Exchange CSW Endpoint" %}</a></li>
               {% if USE_OSGEO_IMPORTER %}
                   <li><a href="{% url "uploads-list" %}"><i class="fa fa-th-list"></i> {% trans "Staged Layers" %}</a></li>
               {% endif %}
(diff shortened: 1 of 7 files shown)


exchange/templates/base.html
```diff
@@ -207,6 +207,9 @@ <h4 class="modal-title" id="myModalLabel">{% trans "Menu" %}</h4>
           </div>
           <div class="modal-body">
             <ul class="list-unstyled">
+              {% if USE_OSGEO_IMPORTER %}
+                  <li><a href="{% url "uploads-list" %}"><i class="fa fa-th-list"></i> {% trans "Staged Layers" %}</a></li>
```
sarasafavi: I'm not sure what the process is for getting a new phrase added to GeoNode's translations - until "Staged Layers" in localized this just won't translate.

sarasafavi:
This has now been rebased to include #19's updates (RC1 changes) to master. As of right now, osgeo_importer will need patched locally to fix port numbers for gsconfig (discussed w Dan in Slack).

sarasafavi:
GeoTIFF support is now fixed in osgeo_importer master: https://issues.boundlessgeo.com:8443/browse/NODE-198

sarasafavi:
GeoTIFF support is now fixed in osgeo_importer master: https://github.com/ProminentEdge/django-osgeo-importer/pull/31#issuecomment-232395747

harts-boundless:
rebasing this had a lot of merge conflicts, but I think I'm about done on it, so nobody go wading into these merge conflicts for a bit, because it'll be a lot of pain for nothing.

harts-boundless:
I separately PR'ed one of the dev shortcut aliases, and another PR for the geonode urls import.

I'm a little worried about numpy versioning because it seems that upstream django-osgeo-importer does not specify a version, and we basically need to test that everything works as intended given that. I'm also concerned about whether the numpy requirement will make things tough as far as the build environment in pip install, hopefully wheels or something will save the day. even then, I'm not sure how well numpy is doing under pypy these days. overall, if we could pull out numpy or make it optional that would be pretty cool in my book.

@sarasafavi some questions about this patch

the biggest change I made is that there were a lot of geonode deps pinned in our requirements.txt as of the time this PR was submitted, now they have been pulled out. that implies the versions changed. so I guess we need to exercise this pretty well. better than I know how to do. can you assist when you have time?

is it ok if I do a new PR? some commits had to be carved up and you lose attribution for them, while others will remain as-is. I'm sorry for that, we can work on something if it's a concern.

What do you think is the implication of setting UPLOADER conditionally on POSTGIS_URL, as I'm seeing current upstream/master doing, vs setting it unconditionally as your PR did? If it is being set conditionally, I introduced a setting of None so we at least get a NoneType error if someone tries to use it in the wrong conditions. It might be better just to set it unconditionally though, do you think so?

can you help explain why this patch adds `'kombu.transport.django'` to `INSTALLED_APPS`?

is it necessary, to the functionality of tihs PR, in settings to set `DEBUG_PROPAGATE_EXCEPTIONS` or to make those tweaks for tastypie PR1376 (`DEBUG = False`, `ALLOWED_HOSTS = ['*']`) ? Are those changes we might want to make as part of another subject, like a celery debugging story? I'm concerned also that these settings might break production, requiring existing deployments to override these variables where they didn't have to before, on pain of things blowing up. But it's more of a dilemma if these are needed for the functionality of django-osgeo-importer. :/ 

In general I found it hard to distinguish where scaffolding ended and code began in this PR, and I think we shouldn't be afraid to either push all these "local" or "personal" scaffolding changes upstream so everyone can use them, or come up with a different system to hold them so they don't end up in version control conflated with some feature.

There are other logging changes I've deleted because Dan introduced a logging config and I don't want them to fight. Hopefully these don't break anything that was intended in this PR, so we'll have to test that.

I think we should take a hard look at the UI on 'staged layers' . Secondarily, I think we should try to ensure that it harmonizes with the rest of the app. If we are inconsistent in a bad way, we should fix this (maybe in a subsequent PR, as long as it's all same release) but if the rest of the app is what's wrong, we need to fix the rest of the app.

harts-boundless:
Please compare with https://github.com/harts-boundless/exchange/tree/NODE163-integrate-osgeo-importer-2

sarasafavi:
Closing - see #60

PR #19, opened 2016-07-05
Node114 registry

BerryDaniel:
Registry
- updated dev/settings.sh for registry dependencies
- added context_processor for registry for it to not display in admin menu in not enabled.
- added custom aggregator templates for registry and elasticsearch, look is same as exchange.
- added hypermap url patterns to exchange
- added the following requirements django-mapproxy, HHypermap, mapproxy, pycsw

Exchange look and feel 
- modified default logo, icon, banner and hyperlink colors for boundless
- added context_processor for map_crs for it to display in remote services page.
- added custom service_register template to display that the remote service needs to be the default_map_crs
- settings now accepts env variables, but has default values if not set.
- added boundless favicon
- added font-exchange with now has an exchange font logo (currently not used)
- modified base.html to now have a direct option to upload or create a document, layer and map.
- removed create layer option in the user menu.
- upgraded django-polymorphic to 0.9.2

Lato, "Helvetica Neue", Helvetica, Arial, sans-serif

Cloud Foundry
- moved settings to cf.py, currently it is not being imported. Need to adjust for commercial and geoint support.


dev/bootstrap.sh
```diff
@@ -161,7 +161,12 @@ gs-dev_init()
 service_setup()
 {
   service elasticsearch restart
+  sleep 10
+  curl -XDELETE 'http://localhost:9200/hypermap/'
```
sarasafavi: `-XDELETE` --> `-X DELETE` ?

BerryDaniel: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-index.html

Just used the example from here :) 

provisioning works.

PR #26, opened 2016-07-19
use setuptools to find subpackages

harts-boundless:
Right now if you `pip install ./exchange` (or the like) the resulting install will not include subpackages such as exchange.settings in the copy installed in site-packages, because the relevant distutils function does not search subdirectories. This is at least one barrier to using the package in-place from a pip install, requiring it to be used from a copy of the source tree.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@
 #########################################################################
 
 import os
-from distutils.core import setup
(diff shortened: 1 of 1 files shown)


harts-boundless:
As an alternative we could stick to distutils and stick to manually enumerating the packages to include, but why bother

PR #27, opened 2016-07-20
fix egg fragments generating pip warnings

harts-boundless:
This patch avoids some warning messages which are probably harmless.

Since I run out of RAM on my little laptop, I need help testing that everything still builds correctly, or alternatively I can test this in a week.


BerryDaniel:
@harts-boundless Thanks

sarasafavi:
Sanity-checked this on a dev install -- LGTM

PR #30, opened 2016-07-20
let boundless URL be set via settings

sarasafavi:
Let e.g., JWICS Exchange instances specify a custom Boundless URL for footer link


harts-boundless:
Very clean +1

PR #37, opened 2016-08-02
Uploaded resources should be private and not downloadable by default

davisc:
This PR support NODE-41.

screen shot 2016-08-02 at 3 07 16 pm


BerryDaniel:
@davisc Shouldn't the template adjustment be included also? Also can we bump this to 1.0.1 alpha?

davisc:
@BerryDaniel I found that this setting is supported in core GeoNode so the template currently handles it.

PR #40, opened 2016-08-15
[WIP] Docker for development

ingenieroariel:
Works for everything except LDAP and GDAL in development.

```
#python-ldap==2.4.25                                                            
#django-auth-ldap==1.2.7
#GDAL==2.0.1
```


BerryDaniel:
@ingenieroariel 

We were not expecting the additional changes outside of the maploom and registry additions. We will need this adjusted so that it is limited to only those items. We already have someone that has been working on the Docker setup for exchange, which will include the GDAL package.

Can you adjust the pull request?

fyi, @davisc

ingenieroariel:
Sure @BerryDaniel - we separated the actual changes away from the Docker part and this can be safely closed.

PR #42, opened 2016-08-18
don't use pickle serialization in celery

harts-boundless:
This is just a proposal and it needs behavioral testing before merge should be considered.

Unpickling can execute arbitrary Python code, which is pretty bad when the pickle is coming over the network, which is the only use that Celery has for pickle. Celery isn't really armored against this in other ways, so normal practice with Celery is to disable pickle serialization, and the Celery project has been struggling to deprecate pickle as a default if not totally disallow it for years now. According to the deprecation warnings, they will finish this pretty soon, which means we will have to ensure we don't rely on pickle if we ever want to upgrade celery.

This needs to be tested thoroughly because once an app relies on pickle serialization, it can sometimes be obscure to find the edge cases where JSON can't represent something that we are trying to send. (The whole point of pickle was to serialize/deserialize totally arbitrary Python objects, where JSON is a bit more restricted). However, for most normal uses of Celery, there is absolutely no need to pass arbitrary Python objects, and JSON is really sufficient.


BerryDaniel:
@harts-boundless can we add this to the docker build and then we can run tests on it?

harts-boundless:
no problem, I'll just include it with that branch

PR #44, opened 2016-08-23
add dev shortcut to start celery worker

harts-boundless:
This was one of Sara's shortcuts from back in June, it's a commit from when she was working django-osgeo-importer integration and was probably useful then for repeatedly bringing the celery worker up and down for testing inside the vagrant environment.

I'm -0 on this particular change, especially given that I'm hoping to make the docker config usable as the default dev environment, but it doesn't really hurt anything. (If we keep all this, I really think though that these aliases should be in their own file instead of appending behind a grep check, that's really a bit goofy)

I just don't want this kind of scaffolding change to appear in feature branches, so if someone wants it to go in, I am doing it as a separate PR to keep things cleaner. But if nobody cares, we'll close it and drop the subject


harts-boundless:
I think I'm going to close this and remove all the aliases in a separate PR. Objections?

PR #47, opened 2016-08-25
remove shell aliases in docker config

harts-boundless:
The aliases are left intact in the Vagrant config, which is not touched.

I originally included these for continuity with the vagrant config, but in the docker config I don't think these are needed at all and are only going to cause confusion.


harts-boundless:
@sarasafavi if you don't use these in docker config let's merge this and blow them away

harts-boundless:
sara seconded this, I'm merging to avoid more administrative overhead

PR #54, opened 2016-08-29
use initial.json instead of initial_data

harts-boundless:
There was a reason why initial_data.json was renamed to initial.json:
initial_data is autoloaded every time migrations are run, which was causing
some data to be lost: in thi scase, it would switch to the default template,
when some deploys for clients had another template.
(Django is deprecating that automagic soon anyway, but we didn't want it here.)

However, we do still want to load that data in a new instance, hence
this commit. Just renaming initial_data to initial left new dev instances in
a dumb state (e.g.: ValueError at /account/login/ - The 'logo_image' attribute
has no file associated with it.)

Unfortunately, ./manage.py loaddata some_invalid_fixture_name actually
runs with exit code 0, so the scripts weren't halting as they should have
in order to highlight the problem, but what can you do

### DIFF
--- a/dev/bootstrap.sh
+++ b/dev/bootstrap.sh
@@ -73,7 +73,7 @@ exchange_setup()
     python /vagrant/manage.py migrate --noinput
     python /vagrant/manage.py collectstatic --noinput
     #hotfix, need to find out why it is not importing the categories
-    python /vagrant/manage.py loaddata initial_data
+    python /vagrant/manage.py loaddata initial
     #echo "from geonode.people.models import Profile; Profile.objects.create_superuser('admin', 'admin@exchange.com', 'exchange', first_name='Administrator', last_name='Exchange')" | python /vagrant/manage.py shell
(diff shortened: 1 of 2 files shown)


harts-boundless:
Erik is getting DoesNotExist at / - Profile matching query does not exist.

BerryDaniel:
@harts-boundless is Erik getting the issue listed above due to the change you have in the PR.

harts-boundless:
I don't know the answer to that until I know why Erik is getting the issue he's getting, if it involves a change in how fixtures are loaded then that is possible

harts-boundless:
OK, I'm content that this issue was due to an existing wrong ordering of migration steps (a matter for subsequent PRs). Feel free to merge as desired

harts-boundless:
However, even as of this commit the app is not working in local dev environments unless you manually run `manage.py migrate account` after boot, which is pretty lame

harts-boundless:
See #58 for a solution to that

PR #61, opened 2016-08-31
Fix CSW-T in MultiCatalog mode

ingenieroariel:
Registry was inserting records in the default catalog instead of the one in the url namespace.


harts-boundless:
As a worker on Exchange, I really appreciate such a small and focused commit. Thank you.

It will also help Exchange maintenance to provide more information on these changes, particularly dependency changes, and particularly information about intent. For example, in this case:

I suppose there might be a JIRA issue somewhere which says that we need to support CSW-T, and maybe even why or in what milestone. If you had a link to such information, that would be great for most people working on Exchange to understand what was going on.

Then I guess that the previously pinned version of pycsw does not support CSW-T, and that the version you are pinning here is the first that does support CSW-T. If there are any other considerations leading to the selection of this version (e.g. if a previous version supported CSW-T but was broken in some way), those would be useful to know too. 

I suppose that pycsw must have no actually released version or even tag that was suitable for us to pin to in order to support CSW-T, so that we were forced to use a commit hash.

I suppose that this was tested, but I don't know how and I wouldn't know (for example) how to test that a given version pin for pycsw properly supported CSW-T. It could be useful to know, if other changes come later on which might cause regressions in CSW-T support.

If you can volunteer a bit more information on these commits to help other people who will be working on Exchange then that will really be appreciated!

ingenieroariel:
Thanks @harts-boundless - I provided more information to @davisc on a private slack channel, including how to test but should have done so here too.

The next pycsw release will be 2.0.1, that development is getting ready to release it and currently making sure it is compatible with an upgraded lxml.

The reason for this pull request is to fix an issue found in the integration of an existing feature in Registry and the MultiCatalog (JIRA-344) functionality added to this version. The bug found was related to the insertion of CSW-T records in the default catalog independently of which catalog it was pushed to. This was fixed in the commit referenced in the PR and will make it to 2.0.1: https://github.com/geopython/pycsw/commit/fdeac838a5e2af0257dbb510445c737fcd5446d2

Here is a sample request to test CSW-T functionality, it contains both what to give and what to expect from it:
https://paw.pt/bGSwq6Z8

screen shot 2016-08-31 at 9 53 03 am


---

I agree with you on the preference to use tags, in fact I helped yesterday get Exchange to the 2.5.2 and pushed to have django-registry for this version taken from PyPi. Unfortunately, Pycsw is an upstream project that we have no control over and will have to wait.

I also noticed the documentation on your Docker pull request and believe it is a high bar worth emulating, thanks for the example. I'll do my best on the PRs coming in the future, even if they are a one liner.

harts-boundless:
Thanks for adding all this supplementary information, it's useful. 

I did a quick smoke test rebased on master and saw no obvious problems.

But I also wasn't sure how to exercise CSW properly through the Exchange UI, so I'll leave that to people who are more qualified on CSW/registry issues

PR #69, opened 2016-09-09
MapLoom update to use swagger api for registry + v1.0.2 updates

BerryDaniel:
Of the 1.0.2 registry updates this will be the only change to master. In the future there should not be any additional updates to MapLoom in regards to registry, since it now pulls data via an api. 

The reason why this will be the only change is that in the next release we will be adding a PR soon to remove registry from exchange and install it separetly in the rpm. A separate branch v1.0.2 has been created that includes the new registry changes. This maploom version has been tested in client environment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 # TODO: Make sure requirements.txt are complete for exchange's env.
 # GeoNode is now using >= and <= in setup.py, if you need specific versions,
 # they should be listed in this file.
-git+git://github.com/boundlessgeo/django-exchange-maploom.git@5722937aba061ddcf3bd2c7b31d6cd616bafbbd9#egg=django-maploom
+git+git://github.com/boundlessgeo/django-exchange-maploom.git@v1.5.10#egg=django-maploom
 git+git://github.com/geopython/pycsw.git@4fc1fc1eee3ac4b83d2862d311b20efb142e13bd#egg=pycsw
 geonode==2.5.2
 dj-database-url==0.4.1


harts-boundless:
This PR looks good to me, runs, and doesn't break anything I can immediately see. So I'm happy to merge it.

As a reviewer, I'm happy to merge version pin changes as frequently as needed as long as we can be reasonably sure they aren't breaking master or preventing us from landing other features we're committed to for the sprint. 

To improve process in the future, I'd like to start asking for PRs with dependency changes to come with clear explanations of which client requirements justify them. Usually all a developer would have to do is refer to a JIRA issue number to satisfy this requirement, assuming requirements have been explained correctly there.

In this case, I find the information in JIRA relating to registry to be unhelpfully vague. Reducing the frequency of Exchange changes due to Registry will be a helpful workaround, but it should never have been the case that Exchange developers were kept in the dark about registry when registry is affecting the client's acceptance of our product, and the same client requirements have driven major changes to Exchange. That would have been the right fix and unfortunately it is not something that Dan and Clarence could unilaterally fix.

Just for reference, here are the JIRA issues I found that seemed to possibly relate.

https://issues.boundlessgeo.com:8443/browse/SEARCH-7 indicates why APIs are being documented and why Swagger is used. This one is clear.

https://issues.boundlessgeo.com:8443/browse/NODE-347 is cryptic, but I guess it could be the core issue, and it mentions "the Swagger API work from GeoNode" - which is suggestive, if not exactly explanatory.

https://issues.boundlessgeo.com:8443/browse/NODE-346 references registry and might reference NODE-347 ("the 3rd party API"). There is also a comment from Jorge Piensa: 'Self documenting Swagger API endpoints have been added and are available at /registry/api/docs'. It's still not clear from this how this involves Exchange, but it seems clearly connected to this PR somehow.

I also don't know what "v1.0.2 updates" means and won't prefer justification for PRs to take forms similar to "Updates for August."

Transparency is not just something that developers give to managers and clients, information should also come to developers.

PR #71, opened 2016-09-12
Added setting to toggle GeoGig workflow

travislbrundage:
https://issues.boundlessgeo.com:8443/browse/NODE-325

### DIFF
--- a/exchange/settings/default.py
+++ b/exchange/settings/default.py
@@ -380,3 +380,6 @@ AVATAR_GRAVATAR_SSL = True
 # CELERY_ACCEPT_CONTENT = ['json']
 # CELERY_TASK_SERIALIZER = 'json'
 # CELERY_RESULT_SERIALIZER = 'json'
+
+# Enable a git-like workflow with geogig
+GEOGIG_WORKFLOW = False


sarasafavi:
@travislbrundage what is this supposed to do?

travislbrundage:
@sarasafavi Originally it was just supposed to be a boolean variable to toggle a GeoGig workflow on or off for an Exchange instance.

The intent was just that as we build the GeoGig workflow to reference this setting, so that the views would be available if `GEOGIG_WORKFLOW = True`.

I'm not sure this is the correct approach anymore and we've been discussing this workflow a lot the past few days internally. I am still working on my understanding of what is appropriate going forward for such a large feature - so maybe we are packaging it up as a django app to install instead.

Following @harts-boundless 's protocol he outlined for Pull Requests, I think it's best I close this for now as I am no longer sure about it.

harts-boundless:
@travislbrundage This looks okay to me, but I think we would want to include the code which consumes this setting in the same PR with the creation of the setting to ensure we don't have superfluous settings. Let me know on Slack DM if I said something that made this more confusing than it had to be.

PR #80, opened 2016-09-21
The standalone TEMPLATE_* settings were deprecated in Django 1.8 and …

BerryDaniel:
…the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DIRS, TEMPLATE_CONTEXT_PROCESSORS, TEMPLATE_DEBUG. https://docs.djangoproject.com/en/1.8/topics/templates/#support-for-template-engines

### DIFF
--- a/exchange/settings/default.py
+++ b/exchange/settings/default.py
@@ -26,8 +26,6 @@ from geonode.settings import *  # noqa
 from geonode.settings import (
     MIDDLEWARE_CLASSES,
     STATICFILES_DIRS,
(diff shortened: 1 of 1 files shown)


harts-boundless:
Looks good to me and seems to work fine. We'll keep an eye out for issues ahead of any release-type activities.

Something like this might also be useful against GeoNode's settings files.

BerryDaniel:
@harts-boundless that was on my agenda

PR #82, opened 2016-09-22
refactor of custom theme to use external app themes

BerryDaniel:
cleanup of some files that were dependent for the single custom theme; html & static img files (now packaged with django-exchange-themes), fixtures (replaced with migrations in themes app).


harts-boundless:
This is working for me.

Not an issue with the PR: due to removal of files like `admin.py` there may be remaining files like `admin.pyc` in directories that are shared with vagrant VMs or docker containers, that may need to be deleted

PR #84, opened 2016-09-26
Add Selenium testing

onwsk8r:


harts-boundless:
Nice upgrade from the curl line. Thanks!

harts-boundless:
For reference, the JIRA issue is NODE-435

PR #92, opened 2016-10-25
Allow for exchange to be installed as a python package

BerryDaniel:
added a manifest file to exclude the cf app (not ready for distribution) and to allow for specific non python file types.

adjusted readme to reStructuredText. Modified readme to exclude the half written steps.

modified version.py to not require geonode app to generate the version. Also removed the git changes, since we do not do dev releases.

modified classifiers, name from exchange to geonode-exchange (exchange was taken on pypi) also made the adjustment from .md to .rst for the readme.


harts-boundless:
I don't see this causing any problems and it could be internally useful (e.g. to let us pip install from an exchange checkout, or use tox).

PR #94, opened 2016-10-26
Use new link to load user activity correctly

travislbrundage:


travislbrundage:
Refer to PR #95 instead

PR #95, opened 2016-10-26
Updated profile detail template for updates to My Activities and Reso…

travislbrundage:
…urces sections

Requires updates from GeoNode, as of https://github.com/GeoNode/geonode/commit/102584f006aba71c1c1c6dc31a7f2455391759a2 on 10/17/16.

### DIFF
--- a/exchange/templates/people/profile_detail.html
+++ b/exchange/templates/people/profile_detail.html
@@ -127,9 +127,7 @@
     <ul class="list-group">
       <li class="list-group-item"><a href="{% url "layer_upload" %}"><i class="fa fa-cloud-upload"></i> {% trans "Upload new layers" %}</a></li>
       <li class="list-group-item"><a href="{% url "new_map" %}"><i class="fa fa-map-marker"></i> {% trans "Create a new map" %}</a></li>
-      {% if_has_tag actor_url %}
-      <li class="list-group-item"><a href="{% actor_url profile %}"><i class="fa fa-fire"></i> {% trans "My Activities" %}</a></li>
-      {% endif_has_tag %}
+      <li class="list-group-item"><a href="{% url "user-activity" profile.username %}"><i class="fa fa-fire"></i> {% trans "My Activities" %}</a></li>
     </ul>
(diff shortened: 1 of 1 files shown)


harts-boundless:
A little more detail will help others understand the change.

What does the update do?

Why does it do that? Is there any applicable client request or discussion which drove the change?

Why does it require a change to that particular version of GeoNode?

travislbrundage:
This change updates the profile_detail template in two ways.

1) Removes a line of angular code, `ng-if = results.length > 0`, since this would improperly consume the entire div when satisfied. It resulted in missing functionality on the profile detail page. This section corresponds to the "Resources" section on the profile detail page, which would inappropriately disappear if *any* of the resource types had no data, which is unwanted behaviour. A detailed bug explanation is in [NODE-4xx].

This update was reported by GVS but affects core GeoNode - its profile_detail template was updated the same way, now just updating ours.

This change does not require any updates from GeoNode, but recent updates will make it look better.

I can pull this one out separately since it's a small, harmless fix that doesn't rely on GeoNode.

2) Updating the activities link to use a new view, present in GeoNode as of https://github.com/GeoNode/geonode/commit/102584f006aba71c1c1c6dc31a7f2455391759a2 (I actually picked the wrong commit earlier; fixed in the OP).

The old user activities view was not working as intended. The activity feed was partially being displayed, but would have bugs where sometimes the layer/map name would not be listed (e.g. reads `user created ____`) and would never have links to the resources. This happened because the user's activity feed was improperly reusing the general Recent Activity feed without passing it proper information. Since it needed to function a slightly different way, a new view was created which would capture the user's username to appropriately create the activity feed.

This requires the new view from GeoNode, otherwise we are linking to a view that does not exist.

This was another bug reported by GVS which affected general GeoNode, and has been updated in GeoNode.

harts-boundless:
That helps, thanks for the additional detail.

harts-boundless:
At casual inspection this isn't breaking anything and I'd like to move ahead, so let's just do this

PR #97, opened 2016-10-27
Exchange release 1.1.0rc1

BerryDaniel:
- Updated version to 1.1.0c1
- Removed gunicorn as a dependency from exchange
- set versions for django-exchange-maploom and django-exchange-themes
- removed TODO section from requirements.txt
- added set versions for 1.1.0rc1 release in the install_requires section of setup.py

### DIFF
--- a/exchange/__init__.py
+++ b/exchange/__init__.py
@@ -23,7 +23,7 @@ try:
 except ImportError:
     pass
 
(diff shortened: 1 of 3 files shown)


harts-boundless:
This seems to cover all different subjects.

There is no explanation of what is changing with the dependency changes.

There is no explanation of why any of this is being done.

PR #99, opened 2016-11-01
tweak ALLOWED_HOSTS so docker can work again

harts-boundless:
The "registry removal" PR broke the docker dev environment.


harts-boundless:
Nothing in this changeset should break Travis CI builds. For proof, see #100 which more literally does nothing and also fails the Travis CI build, but does not involve any of the changes in this PR here.

To see recent commit history, here's a link: https://github.com/boundlessgeo/exchange/commits/master. There are only a few commits in the past week.

Either the build failures are non-deterministic, some recent condition arose which generates false positives for e.g. #99 and #100, or earlier Travis CI builds passed commits which shouldn't have passed (false negative).

PR #102, opened 2016-11-02
default osgeo-importer handlers now in settings

sarasafavi:
In django-osgeo-importer, default handlers have been moved out of code and into project-level settings to (ultimately) support geoserver-independent geonode layer imports.

_Reference https://github.com/boundlessgeo/django-osgeo-importer/pull/43 for details_


harts-boundless:
importer went ahead with this because mapstory had already done the needful here, so exchange was the only one left. There should be an issue filed soon about figuring out and documenting how to use importer without geoserver present, however that may not apply to exchange any time soon.

sarasafavi:
@harts-boundless relevant issue now at https://github.com/GeoNode/django-osgeo-importer/issues/58

harts-boundless:
This is absolutely required for importer HEAD to work in Exchange. Looks good.

PR #103, opened 2016-11-05
Remove /services endpoint temporarily

travislbrundage:
Removing any access to registry services. Currently using the registry services can cause errors in MapLoom, so we've elected to remove the endpoint for the time being so clients cannot use registry services and cause the bad state on their Exchange instances.

### DIFF
--- a/exchange/templates/500.html
+++ b/exchange/templates/500.html
@@ -1,0 +1,308 @@
+{% load i18n avatar_tags %}
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1"> 
+    <title>{% block title %}{{ SITE_NAME }}{% endblock %}</title>
+    {% block head %}
+      {% if DEBUG_STATIC %}
+      <link href="{{ STATIC_URL }}lib/css/jquery.dataTables.css" rel="stylesheet" />
+      <link href="{{ STATIC_URL }}lib/css/select2.css" rel="stylesheet"/>
+      <link href="{{ STATIC_URL }}lib/css/bootstrap.min.css" rel="stylesheet"/>
+      <link href="{{ STATIC_URL }}lib/css/jquery-ui.css" rel="stylesheet"/>
+      <link href="{{ STATIC_URL }}lib/css/bootstrap-datetimepicker.css" rel="stylesheet"/>
+      {% else %}
+      <link href="{{ STATIC_URL }}lib/css/assets.min.css" rel="stylesheet"/>
+      {% endif %}
+      <link href="{{ STATIC_URL }}geonode/css/ext-compatibility.css" rel="stylesheet" />
+      <link href="{{ STATIC_URL }}geonode/css/base.css" rel="stylesheet" />
+      <style type="text/css">[ng\:cloak],[ng-cloak],[data-ng-cloak],[x-ng-cloak],.ng-cloak,.x-ng-cloak,.ng-hide:not(.ng-hide-animate){display:none !important;}</style>
+      {% block extra_head %}
+      {% endblock %}
+    {% endblock %}
+    <!--[if IE]>
+      <script src="https://html5shim.googlecode.com/svn/trunk/html5.js"></script>
+      <style>
+        #ieflag{
+            display: block !important;
+            height: 150px;
+            text-align: center;
+        }
+      </style>
+    <![endif]-->
+  </head>
+  
+  <body class="{% block body_class %}{% endblock %}">
+
+    {% block header %}
+    <nav class="navbar navbar-inverse navbar-fixed-top" role="navigation">
+      <div class="container">
+        <div class="navbar-header">
+          <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar">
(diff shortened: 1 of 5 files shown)


harts-boundless:
Can you help explain some things about this? 

I don't know what the goal or context is. If there is documentation somewhere else to explain that, can you link to it? While the commit messages says this removes `/services`, this appears to actually add a new route with a redirect. I gather that this is intended to deny access to the `/services` URL, but I don't know why this is being done or why this particular approach was chosen. Here is where I am hoping that just a little bit of communication of the reasoning will allow me to understand or be more helpful in finding a solution to the problem you're working on.

One of my approach questions is about the relationship to GeoNode here. You mention registry services, but unless I'm mistaken, this URL is actually coming from GeoNode (geonode/urls.py line 108). Does GeoNode depend on registry, or include it by default? What does hitting this URL do in vanilla GeoNode? Can you explain why we are dropping support for routes that are provided by vanilla GeoNode? Maybe this is just me, but I would hope to reduce the differences between Exchange and GeoNode, not increase them. If GeoNode is going to lose these routes, I think it would be OK for Exchange to have them until GeoNode loses them. But if GeoNode is not going to lose these routes, then I would expect to retain them in Exchange, as a flavor of GeoNode.

My other approach question is about the choice of a redirect. I don't know that adding a redirect is a great way of handling a URL that should not exist (assuming that for some reason this URL should not exist). If `/services/` should not exist in Exchange for some reason, a 404 makes the situation much clearer than adding a redirect to the index. It should be sufficient for us not to have any links to anything we don't want users hitting. It is not a real issue if a user manually types in a URL and gets a 404, is it?

harts-boundless:
Even if this is merged, there are still going to be a number of links which point to this URL; while this PR will "disable" them and so in that sense users might learn not to use them, those links will still be present in the UI to confuse people.

I went to find some of the links we would have to disable to cleanly remove `/services` from user view.

There are a few GeoNode templates that link to some URL under `/services/`, e.g.

in GeoNode `geonode/templates/base.html` (in Exchange this is: `django-exchange-themes` - `appearance/templates/base.html`)
```
    {% if user.is_superuser %}
    <li><a href="{% url "services" %}"><i class="fa fa-globe"></i> {% trans "Remote Services" %}</a></li>
    ...
```

in GeoNode, `geonode/templates/500.html`
-- this file substantially duplicates `base.html`, including the above snippet.
-- Exchange probably inherits this from GeoNode

`geonode/layers/templates/layers/layer_detail.html`
-- hardcodes "/services/..."
```
    {% if resource.storeType == "remoteStore" %}
    <li class="list-group-item">
            <h4>{% trans "External service layer" %}</h4>
            <div>{% trans "Source" %}: <a href="/services/{{resource.service.id}}">{{ resource.service.title }}</a></div>
```
in Exchange, `exchange/templates/people/profile_detail.html`
```
      {% if not REGISTRYURL %}
      <li class="list-group-item"><a href="{% url "services" %}"><i class="fa fa-globe"></i> {% trans "Remote Services" %}</a></li>
      {% endif %}
```

There are several layer views which may interact with 'services' if they are added (via `geonode.layers.views._resolve_layer`, which is called from a ton of views, and itself calls `geonode.utils.resolve_object`). 

Then there is the admin, which may well still allow adding services, if someone was foolish enough to seek it out. So we should probably find a way to disable that too if our intention is to prevent people from adding services in the GeoNode sense.

We may also want to remove `exchange/templates/services/service_register.html` if we aren't exposing that page, and therefore don't need to customize that GeoNode template any more.

If GeoNode isn't going to use `/services` any more, then it should be possible to clean out `geonode/services/*` but I have a hunch that is not where this is going

travislbrundage:
So essentially what I understand is that the registry services are causing errors in MapLoom. Layers that are added this way don't work correctly. In order to prevent our users from entering this bad state, we're doing a quick fix to disable this endpoint.

I agree with you that a 404 makes more sense. I will be updating and resubmit this pull request with that change made.

@davisc Can you confirm that is correct or explain why this change is needed?

harts-boundless:
It sounds to me like MapLoom is the best place for these changes then. If there is an error in MapLoom, we need to understand the cause of that error, and document it here. Maybe MapLoom should not be hitting /services at all. Or maybe it should only be hitting it on certain conditions (e.g. REGISTRYURL is defined or something). Or maybe Django shouldn't be passing /services as a link to MapLoom, if it is doing that (again, this may depend on things like REGISTRYURL). Or maybe Django needs to pass something like REGISTRYURL to MapLoom so MapLoom can behave appropriately.  Or maybe MapLoom should detect/handle this kind of error more gracefully. It's hard to know without reproducing the issue locally and digging into MapLoom with it.

But the only reason to make a change to /services itself - as a way of changing MapLoom behavior - would be that we literally can't correct MapLoom (or how we talk to it). 
So I'm -1 on changing any GeoNode routes here. Until these routes are deprecated in GeoNode, I think they should be working in Exchange in any case, but certainly if Registry is not in use. 

And again, if we do "disable" /services, there are still all these links to that URL which users will see, and now we are giving them links to a 404, unless we remove those links from our templates. So even if we want to merge this change, I think that should come with removal of the newly dead links in the same PR. But that means even more template departures from GeoNode and I don't see why we would go through that headache if we can fix the issue properly through MapLoom. 

2c.

travislbrundage:
Discussing with Clarence more. Will reopen with adjustments and explanations when finished.

davisc:
Have we confirmed that with this change, the error condition can be avoided if a remote server is added via maploom?

PR #104, opened 2016-11-07
Install GeoNode deps from requirements for development

sarasafavi:
Use dependency versions explicitly pinned in GeoNode's requirements.txt for development. See https://github.com/GeoNode/geonode/issues/2700 for related issue


harts-boundless:
Ironically, this is running up against the same problem it is trying to fix (Travis CI is now broken because of the same issue).

harts-boundless:
To clarify, this fix only applies to the docker-compose dev env, but it is the "right" fix - it uses all of GeoNode's exact version pins.

PR #106, opened 2016-11-20
Added a break out for MapLoom

theduckylittle:
Lets a docker based dev environment use a development version of MapLoom

### DIFF
--- a/.env
+++ b/.env
@@ -2,3 +2,6 @@
 GEONODE_HOME=~/boundless/geonode
 # Host path to mount inside container as /mnt/exchange
 EXCHANGE_HOME=~/boundless/exchange
+
+# Host path to mount inside the container as /mnt/maploom
+MAPLOOM_HOME=~/boundless/MapLoom
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,5 +80,12 @@ ADD docker/home/* /root/
 RUN mv /root/local_settings.py /env/lib/python2.7/site-packages && \
     mv /root/settings.sh /etc/profile.d/
 
+# this will symlink the maploom files to the MapLoom repository which
+#  exists outside of the container.
+RUN rm -rf /env/lib/python2.7/site-packages/maploom/static/maploom && \ 
+    ln -s /mnt/maploom/build /env/lib/python2.7/site-packages/maploom/static/maploom && \
+    rm /env/lib/python2.7/site-packages/maploom/templates/maps/maploom.html && \
+    ln -s /mnt/maploom/build/maploom.html /env/lib/python2.7/site-packages/maploom/templates/maps/maploom.html
+
 WORKDIR /scratch
 CMD ["/root/exchange.sh"]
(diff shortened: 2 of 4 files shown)


harts-boundless:
I think we should do whatever is required to make the frontend workflow more comfortable. I wish I had more time to help with this, but we can merge it as-is and fix things up later if you need it.

Can the work be done in the startup script instead of `Dockerfile`?

Can we use `pip install -e /mnt/django-exchange-maploom` instead of directly putting symlinks into the virtualenv? Then in the worst case if you have to do symlink magic, it is in `/mnt/django-exchange-maploom` instead of directly in the virtualenv, or I'm not sure, maybe the symlinks can even be managed in the host. 

Basically I am trying, under the constraint of not increasing the daily friction for frontend work, to reduce the rate at which all the magical host mount stuff does confusing things e.g. because of garbage files that arise in working directories, or because of forgetting what branch you had a share on ... I have already contemplated getting rid of all the mounts altogether because of the problems they have caused.

harts-boundless:
Or maybe `django-exchange-maploom` could use a submodule or something? Just throwing out ideas

theduckylittle:
***WARNING!*** MapLoom will need to be built after cloning! Otherwise you are likely to be shown the following error:

```
 File "/mnt/exchange/manage.py", line 30, in <module>
   execute_from_command_line(sys.argv)
 File "/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
   utility.execute()
 File "/env/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
   self.fetch_command(subcommand).run_from_argv(self.argv)
 File "/env/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
   self.execute(*args, **cmd_options)
 File "/env/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
   output = self.handle(*args, **options)
 File "/env/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 168, in handle
   collected = self.collect()
 File "/env/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 107, in collect
   handler(path, prefixed_path, storage)
 File "/env/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 305, in copy_file
   if not self.delete_file(path, prefixed_path, source_storage):
 File "/env/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 255, in delete_file
   self.storage.delete(prefixed_path)
 File "/env/lib/python2.7/site-packages/django/core/files/storage.py", line 288, in delete
   os.remove(name)
OSError: [Errno 21] Is a directory: '/scratch/static_root/maploom'
```

To build MapLoom (assumes npm and bower are installed):
```
$ cd ~/boundless/MapLoom
$ npm install ; bower install ; grunt build
````

PR #107, opened 2016-11-22
bumped versions for documentation and themes to now include the new l…

BerryDaniel:
…ook and feel regarding logos. I also modified the activate script in vagrant dev to be a +e, which will prevent the auto kickout of a login after a failed command.


BerryDaniel:
@davisc Can this be merged?

PR #115, opened 2016-12-12
Disabling registry services by default

travislbrundage:
This is a redo of PR #103 which should prevent the user from being able to use registry services by default in Exchange. They can be enabled by setting the `REGISTRY` variable to `True`.

Although it is not visible, this same variable is present in MapLoom. By setting it to False, it should prevent the error state from being reachable in MapLoom as well.

Therefore:
- Setting the REGISTRY variable in Exchange's settings to False prevents a potential error state in MapLoom
- We can also leverage the REGISTRY settings variable to block the endpoint in Exchange, so the user cannot use functionality that they won't be able to access in MapLoom anyway

### DIFF
--- a/exchange/settings/default.py
+++ b/exchange/settings/default.py
@@ -375,7 +375,7 @@ if GEOAXIS_ENABLED:
 
 
 # NEED TO UPDATE DJANGO_MAPLOOM FOR ONLY THIS ONE VALUE
-REGISTRY = os.environment.get('ENABLE_REGISTRY', False)
+REGISTRY = os.environ.get('ENABLE_REGISTRY', False)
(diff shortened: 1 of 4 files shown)


exchange/settings/default.py
```diff
@@ -366,6 +366,7 @@ def str2bool(v):
 
 
 # NEED TO UPDATE DJANGO_MAPLOOM FOR ONLY THIS ONE VALUE
+REGISTRY = False
```
davisc: Lets grab the value from the os.environ and set the default to False.

Example:
Assumes 'ENABLE_REGISTRY' is added to the environ.
os.environment.get('ENABLE_REGISTRY', False)

PR #116, opened 2016-12-14
configure settings file to support all env variables

BerryDaniel:
cleaned up the settings file to allow for environment variables. I did leave out osgeo_importer, to prevent any accidental deployments. The default setting for exchange is to use postgres and postgis for django, geoserver and geogig. There is no longer an option for sqlite, since we do not use it and if the settings are not properly configured it will display that in the logs rather than just creating a sqlite db. I did add the option for classification, needed for dod customers and registration / email server needed for the trial and demo server.

### DIFF
--- a/exchange/settings/default.py
+++ b/exchange/settings/default.py
@@ -36,43 +36,76 @@ from geonode.settings import (
 def str2bool(v):
     return v.lower() in ("yes", "true", "t", "1")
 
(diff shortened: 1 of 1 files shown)


BerryDaniel:
@davisc Can you review and merge. We need this for updating the demo and trial chef scripts.

PR #118, opened 2016-12-15
added geoaxis authentication middleware to exchange core. 

BerryDaniel:
This will limit the amount of packages we will have to maintain and get accredited.

To enable add an os environment for GEOAXIS_ENABLED=True

Note: needed for multiple customer projects

@davisc please review. This will allow us to delete django_webgate and just use this middleware.


exchange/auth/geoaxis.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from django.contrib.auth.middleware import RemoteUserMiddleware
+
+
+class OamRemoteUserMiddleware(RemoteUserMiddleware):
```
davisc: I think we should follow the same convention as django. 

`exchange/auth/middleware.py`
`class GeoAxisMiddleware(RemoteUserMiddleware)`

exchange/settings/default.py
```diff
@@ -358,6 +360,13 @@ def str2bool(v):
     AUTH_LDAP_USER_SEARCH = LDAPSearch(LDAP_SEARCH_DN,
                                        ldap.SCOPE_SUBTREE, AUTH_LDAP_USER)
 
+# geoaxis
+GEOAXIS_ENABLED = str2bool(os.getenv('GEOAXIS_ENABLED', 'False'))
+if GEOAXIS_ENABLED:
+    AUTHENTICATION_BACKENDS = (
+        'exchange.auth.geoaxis.OamRemoteUserMiddleware',
```
davisc: Based on the comment above, this would change to...

`exchange.auth.middleware.GeoAxisMiddleware`

exchange/settings/default.py
```diff
@@ -358,6 +360,13 @@ def str2bool(v):
     AUTH_LDAP_USER_SEARCH = LDAPSearch(LDAP_SEARCH_DN,
                                        ldap.SCOPE_SUBTREE, AUTH_LDAP_USER)
 
+# geoaxis
+GEOAXIS_ENABLED = str2bool(os.getenv('GEOAXIS_ENABLED', 'False'))
+if GEOAXIS_ENABLED:
+    AUTHENTICATION_BACKENDS = (
```
davisc: Need to make sure 'django.contrib.auth.middleware.AuthenticationMiddleware' is added before the RemoteUserMiddleware class.

PR #120, opened 2016-12-20
Set a limit for the text length

BerryDaniel:
This will prevent the title from running into the next item. Modified from h4 to h5 to decrease the text size. This now allows for 24 characters.

Also bumped GeoNode version to include the dependency of corsheaders app.

text_width


exchange/themes/templates/index.html
```diff
@@ -187,7 +187,7 @@
           <div ng-repeat="item in featured | limitTo:4">
             <div class="col-md-3">
               <a href="{{ item.detail_url }}"><img ng-src="{{ item.thumbnail_url }}" /></a>
-              <h4>{{ item.title }}</h4>
+              <h5>{{ item.title | limitTo: 24 }}{{item.title.length > 24 ? '...' : ''}}</h4>
```
davisc: Check out the built in template tag for truncatechars and truncatewords.

**truncatechars**

https://docs.djangoproject.com/en/dev/ref/templates/builtins/#truncatechars

Truncates a string if it is longer than the specified number of characters. Truncated strings will end with a translatable ellipsis sequence (”...”).

Argument: Number of characters to truncate to

For example:

`{{ value|truncatechars:9 }}`
If value is "Joel is a slug", the output will be "Joel i...".


**truncatewords**

https://docs.djangoproject.com/en/dev/ref/templates/builtins/#truncatewords

Truncates a string after a certain number of words.

Argument: Number of words to truncate after

For example:

`{{ value|truncatewords:2 }}`

If value is "Joel is a slug", the output will be "Joel is ...".

Newlines within the string will be removed.

PR #123, opened 2016-12-21
Removed attributes from default.py to align with upstream 

amirahav:
The STATIC_ROOT, STATIC_URL, MEDIA_ROOT and MEDIA_URL attributes are already exposed as environment variables in Geonode, and adding them to Exchange causes several issues, including thumbnails being saved in incorrect places (see NODE-624)

This removes those attributes from default.py and adds them to settings.sh.
Note - haven't been tested with Docker.


dev/settings.sh
```diff
@@ -14,4 +14,8 @@ export GEOSERVER_URL='http://'"${IP}"'/geoserver/'
 export GEOSERVER_DATA_DIR='/vagrant/dev/.geoserver/data'
 export GEOSERVER_LOG='/vagrant/dev/.geoserver/data/logs/geoserver.log'
 export GEOGIG_DATASTORE_DIR='/vagrant/dev/.geoserver/data/geogig'
+export STATIC_ROOT = '/vagrant/.storage/static_root'
+export STATIC_URL = '/static/'
+export MEDIA_ROOT = '/vagrant/.storage/media'
+export MEDIA_URL = '/media/'
```
sarasafavi: These spaces around `=` break the script

sarasafavi:
Fixing this for both Vagrant & Docker dev envs (see also comment above); will test, then update branch & merge once I'm done

dev/settings.sh
amirahav: D'oh! and thanks for fixing those!

docker/home/settings.sh
```diff
@@ -17,8 +19,19 @@ export GEOSERVER_URL='http://172.16.238.2/geoserver/'
...
 
 # debug for local dev
 export DJANGO_LOG_LEVEL=DEBUG
 export REGISTRYURL='http://registry.boundlessgeo.io'
+export STATIC_ROOT='/scratch/static_root'
+export STATIC_URL='/static/'
+export MEDIA_ROOT='/scratch/media_root'
+export MEDIA_URL='/media/'
+export DEBUG_STATIC='False'
+export ALLOWED_HOSTS='["172.16.238.3", "exchange"]'
```
sarasafavi: This is duplicating settings vars (STATIC_ROOT, MEDIA_ROOTE, ALLOWED_HOSTS) from https://github.com/boundlessgeo/exchange/blob/master/docker/home/local_settings.py

amirahav: Let me remove those then. Other than that are you seeing any other issues? 
This commit should have fixed the postgres issue with Docker (And was working over the weekend when I tested it)

sarasafavi: @amirahav still testing locally (talked to @BerryDaniel a few minutes ago about same) -- will update here once I'm done

exchange/settings/default.py
```diff
@@ -36,25 +36,13 @@
 def str2bool(v):
     return v.lower() in ("yes", "true", "t", "1")
 
-DEBUG = TEMPLATE_DEBUG = str2bool(os.getenv('DEBUG', 'True'))
-DEBUG_STATIC = str2bool(os.getenv('DEBUG_STATIC', 'False'))
-SITEURL = SITE_URL = os.getenv('SITE_URL', 'http://127.0.0.1:8000')
```
sarasafavi: Since SITE_URL is going away, there's a couple remaining instances of the old var that need find & replaced (cf settings, travis settings, couple more docker-related instances)

Dockerfile
```diff
@@ -3,7 +3,7 @@
 # it isn't meant to be; it's for the benefit of people hacking on Exchange.
 
 # The django app is tested on CentOS 6.7 to match client production environment
-FROM centos:6.7
+FROM centos:6.8
```
sarasafavi: cc @BerryDaniel (who is verifying client's current version) - I'd rather notbump our dev env unless it mirrors a production bump too...

docker/geoserver/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM tomcat:9-jre8
 
 # Grab the WAR file containing the version of GeoServer Exchange depends on.
 # Avoid ADD so we can reuse this layer instead of downloading at every build.
-RUN wget --directory-prefix /tmp --progress=dot:giga https://s3.amazonaws.com/boundlessps-public/GVS/geoserver.war
+RUN wget --directory-prefix /tmp --progress=dot:giga https://yum-geonode.boundlessps.com/war/latest/geoserver.war
 
 # Run config inside container
 RUN unzip -o /tmp/geoserver.war -d /tmp/geoserver && \
```
BerryDaniel: need to add the following

```
    sed -i.bak 's@<onlineResource>http://geoserver.org</onlineResource>@<onlineResource>http://geoserver.org</onlineResource><proxyBaseUrl>http://172.16.238.2/geoserver/</proxyBaseUrl>@' \
               /tmp/geoserver/data/global.xml && \
```
after line 10

Dockerfile
BerryDaniel: Awaiting confirmation from GVS team

BerryDaniel: Looks like they are running 6.6, but geoint would be running the newest 6.8 if an image would be created.

BerryDaniel:
Looks good @amirahav Thanks for the contribution and thanks @sarasafavi for the peer review

PR #124, opened 2016-12-22
Create UI to insert CSW Records

sarasafavi:
Ref: NODE-506 & related subtasks.

This adds a user-facing form for creating new CSW Records in Registry. Form fields are based on preliminary CNP metadata fields (provided by @davisc ) and will likely be modified at a later date.

As this is an "MVP/demo" feature only at this point a direct link has not been added to the user menu. The form is available at `/csw/new` (i.e., 172.16.238.2/csw/new/ in dev); status page is at `/csw/status`.

**Screenshots:**

insert_csw

status

status_updated

status_bad

### DIFF
--- a/exchange/urls.py
+++ b/exchange/urls.py
@@ -41,7 +41,8 @@ urlpatterns = patterns(
         name='geoserver_reverse_proxy'),
     # Redirect help and developer links to the documentation page
     url(r'^help/$', views.documentation_page, name='help'),
-    url(r'^developer/$', views.documentation_page, name='developer')
+    url(r'^developer/$', views.documentation_page, name='developer'),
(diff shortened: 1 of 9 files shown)


exchange/urls.py
```diff
@@ -41,7 +41,10 @@
         name='geoserver_reverse_proxy'),
     # Redirect help and developer links to the documentation page
     url(r'^help/$', views.documentation_page, name='help'),
-    url(r'^developer/$', views.documentation_page, name='developer')
+    url(r'^developer/$', views.documentation_page, name='developer'),
+    url(r'^csw/new/$', views.insert_csw, name='insert_csw'),
```
davisc: Making a note: We should prevent registration of these routes when Registry is not enabled.

Issue #127, opened 2016-12-29
Missing CSS Folder on pip install


benhosmer:
The commit that exchange is using from [osgeo_importer](https://github.com/GeoNode/django-osgeo-importer) doesn't seem to have a `/css` folder:

```
copying osgeo_importer/static/osgeo_importer/Makefile -> build/lib/osgeo_importer/static/osgeo_importer
    error: can't copy 'osgeo_importer/static/osgeo_importer/css': doesn't exist or not a regular file
```

If I change this line in the `requirements.txt`:

```
24 git+git://github.com/boundlessgeo/django-osgeo-importer@b77ea961ea458b9c2ed8388ce7011a43e29f6fe9#egg=    django-osgeo-importer
```

To point to latest, `pip install -r requirements.txt` completes without any errors.

BerryDaniel:
osgeo_importer is still in development and as of now is not included in any production version of exchange. We are currently working on cleaning up the requirements files to allow for a dev (osgeo_importer included) and prod. https://github.com/boundlessgeo/exchange/pull/123

benhosmer:
@BerryDaniel I just fetched from HEAD and this issue no longer exists. Unless you think it should remain open, I'll close it.

PR #128, opened 2017-01-04
setup.py/requirements.txt standardization

amirahav:
In an attempt to make packaging easier all the requirements have been removed from setup.py and moved to requirements.txt

osgeo-importer was moved from requirements.txt to Docker only because it's a dev dependency at the moment, and the commit was changed to solve https://github.com/boundlessgeo/exchange/issues/127

Version bumps for django-exchange-maploom, waitress, supervisor.
Minor version bump for celery to align with upstream (https://github.com/GeoNode/geonode/blob/1898779e8cdf104ac454bc497c6243d19f7caf90/requirements.txt#L12), and could possibly be removed entirely.


requirements.txt
```diff
@@ -16,9 +16,8 @@ psycopg2==2.6.1
 python-ldap==2.4.25
 django-auth-ldap==1.2.7
 GDAL==2.1.0
-supervisor==3.2.3
+supervisor==3.3.1
 python-resize-image==1.1.10
 django-flat-theme==1.1.3
 django-exchange-docs==1.1.2
 flake8==3.0.4
-git+git://github.com/boundlessgeo/django-osgeo-importer@b77ea961ea458b9c2ed8388ce7011a43e29f6fe9#egg=django-osgeo-importer
```
BerryDaniel: @sarasafavi did you want this to stay here or only in docker setup?

BerryDaniel: I think this will ensure that, plus you still need to manually add it to the INSTALLED_APPS.

I will merge it now.

PR #129, opened 2017-01-04
Unified Search

bitner:
Adds unified search using GeoNode Haystack and Registry located on the same elasticsearch instance.
Requires HAYSTACK_SEARCH to be setup and set to True, ES_UNIFIED_SEARCH = True and ES_URL and REGISTRY_URL to be set.

### DIFF
--- a/exchange/views.py
+++ b/exchange/views.py
@@ -370,11 +370,12 @@ def unified_elastic_search(request):
                 # Get source link from Registry
                 xml = value['xml']
                 js = '%s/%s' % (settings.REGISTRY_URL,
-                                re.sub(r"xml$","js",xml))
+                                re.sub(r"xml$", "js", xml))
                 try:
                     req = requests.get(js)
                     json = req.json()
-                    result['registry_url'] = json.sources.default_source.req.url
+                    surl = json.sources.default_source.req.url
+                    result['registry_url'] = surl
                 except:
                     result['registry_url'] = js
             else:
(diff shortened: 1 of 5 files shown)


exchange/views.py
```diff
@@ -127,6 +127,7 @@ def geoserver_reverse_proxy(request):
     return HttpResponse(req.content, content_type='application/xml')
 
 
+<<<<<<< 4d763fbbd0ea00c0dc23c983c1a6a7426eb9a57e
```
BerryDaniel: @bitner not sure how this even passed pep8, looks like you have a merge conflict artifacts L172 and L174 also.

requirements.txt
```diff
@@ -22,3 +22,4 @@ django-flat-theme==1.1.3
 django-exchange-docs==1.1.2
 flake8==3.0.4
 git+git://github.com/boundlessgeo/django-osgeo-importer@b77ea961ea458b9c2ed8388ce7011a43e29f6fe9#egg=django-osgeo-importer
+elasticsearch-dsl==5.0.0
```
BerryDaniel: @bitner is this a hard requirement? If so we need to get this to GEOINT ATO team asap <-- @davisc

bitner: I could do without just using raw elasticsearch, that lib just makes easier

exchange/views.py
bitner: argh, sorry ran flake8 before rebase

requirements.txt
davisc: @BerryDaniel lets update the spreadsheet to include this lib. @bitner can you resolve the conflict below?

PR #131, opened 2017-01-05
Hide password change when remote authentication is enabled

travislbrundage:


BerryDaniel:
@sarasafavi @davisc @bitner @theduckylittle @arhote @staceyberry @onwsk8r 

Just fyi, this is an addition that if GEOAXIS is enabled it will hide the password change part of the template.

PR #132, opened 2017-01-05
Hide logout when remote authentication is enabled

travislbrundage:


BerryDaniel:
@sarasafavi @davisc @bitner @theduckylittle @arhote @staceyberry @onwsk8r 

Another one but this hides logging out, since GEOAXIS will log them back in after logout.

PR #136, opened 2017-01-06
Fix for mysterious "background" layer in MapLoom.

theduckylittle:
GeoNode added a new "no background" layer that is understood
by Geoexplorer but not MapLoom.  It creates an internal "blank"
layer in GeoExplorer but as MapLoom is concerned with remote layers,
it doesn't fit into its data-model. IE, the concept of a "serverless
layer".

This change overrides the default map-background settings from GeoNode and
prevents the otherwise unnamed 'background' layer from showing up
when creating a new map.


BerryDaniel:
Instead of overwriting the `MAP_BASELAYERS` setting for one layer, wouldn't it be best to just remove the blank layer?

theduckylittle:
I was going back-and-forth on that.  We are already making some assumptions by overriding the source of the assumed "0" layer.  And if this 'standard' was established it might give us some more options when/if a client site is required to be in a projection other than Web-Mercator. 

I can otherwise re-write this to do a find/remove.  Just let me know.

theduckylittle:
Closing PR in order to try and fix this in MapLoom.

PR #138, opened 2017-01-09
Force usernames lowercase

travislbrundage:
Forcing usernames lowercase in order to recognize capitals and lowercase as the same user.

i.e. cdavis = CDavis

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,8 @@
 git+git://github.com/boundlessgeo/django-exchange-maploom.git@v1.5.12#django-exchange-maploom
 git+git://github.com/boundlessgeo/dj-database-url@8a24aab17db8ee52400f2fc0fe23c44fef1439e6#egg=dj-database-url
 git+git://github.com/GeoNode/geonode@1898779e8cdf104ac454bc497c6243d19f7caf90#egg=geonode
+# Perhaps make a fork in boundlessgeo and push this branch there, point to it instead
+git+git://github.com/travislbrundage/geonode-user-accounts@force-lower#egg=geonode-user-accounts
 celery==3.1.18
 django-storages==1.1.8
 boto==2.38.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -3,6 +3,8 @@
 git+git://github.com/boundlessgeo/django-exchange-maploom.git@v1.5.12#django-exchange-maploom
 git+git://github.com/boundlessgeo/dj-database-url@8a24aab17db8ee52400f2fc0fe23c44fef1439e6#egg=dj-database-url
 git+git://github.com/GeoNode/geonode@1898779e8cdf104ac454bc497c6243d19f7caf90#egg=geonode
+# Perhaps make a fork in boundlessgeo and push this branch there, point to it instead
+git+git://github.com/travislbrundage/geonode-user-accounts@force-lower#egg=geonode-user-accounts
```
travislbrundage: Requires a modification to this app... should we make a fork in boundlessgeo like I noted and I'll push my changes there?

davisc: @travislbrundage looks like the signup form found in https://github.com/GeoNode/geonode-user-accounts it executes a case insensitive look up of the username via `__iexact` and prevents cDavis from being registered if cdavis exists, however, logging in only honors cdavis and states that cDavis doesn't exist. We may need to look into where the login occurs or create a script to lowercase username for existing data, otherwise these changes work fine.

Before merging this PR, You should submit a PR against https://github.com/GeoNode/geonode-user-accounts or https://github.com/pinax/django-user-accounts as this would be an acceptable change and then update the reference in the requirements.txt.

travislbrundage: I found a different way to do this through a change to GeoNode. It will accept any case for logging in, and match to any case registered on the backend.

ie you can login to cDavis with: cdavis, CDAVIS, Cdavis, all will work.

The registration in the geonode-user-accounts app already had blocked the same username from being able to be registered, so the GeoNode change should suffice.

New PR here: https://github.com/GeoNode/geonode/pull/2821

Thanks!

PR #148, opened 2017-01-24
Use boundless-vendor-libs rpm and remove upstream requirements.txt

BerryDaniel:
The boundless-vendor-libs has been added to docker and vagrant configuration. The location of the libs are now in `/opt/boundless/vendor/lib`, the files have also been adjusted to add postgresql installed at `/opt/boundless/vendor/{lib,bin}`. The inclusion of the vendor-libs allowed for epel to be removed as a dependency. A settings variable, `GEOS_LIBRARY_PATH`, was added for docker config, since a custom location was used for geos.  

There was a change from having the Dockerfile for exchange install all requirements and then have common.sh upgrade the dependencies by using the upstream requirements. A change recently to upstream was causing issues when installing the dependency for six and paver. The change that has been applied to docker setup is to remove the geonode entry in the exchange requirements.txt and then install the local repo at `/mnt/geonode` in common.sh. This also aligns the python package versions that are installed in our rpm and in the vagrant dev setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ install:
   # Install Java, Jetty and Geoserver
   - sudo apt-get -qq install openjdk-8-jdk-headless
   - wget --directory-prefix=/tmp -q http://central.maven.org/maven2/org/eclipse/jetty/jetty-runner/9.3.8.v20160314/jetty-runner-9.3.8.v20160314.jar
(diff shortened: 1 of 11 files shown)


bitner:
Everything appears to pull up and work as expected with both Docker and Vagrant.

PR #151, opened 2017-01-24
Fly-by 'Error' Messages in resource lists.

theduckylittle:
The primary cause of this bug was that the message was showing
before Angular had the chance to parse the document.

Adding "ng-hide" to the div's class makes it invisible to the user
while the page was loading, and then "ng-show" is used to remove
the "ng-hide" class when applicable. But as that is done by the
Angular parser, the content is ready when the div is shown.

Refers to NODE-583


BerryDaniel:
tested and looks good

PR #153, opened 2017-01-24
Use local .env for configuration

stirringhalo:
Makes it easier to experiment with configuration


BerryDaniel:
#202 replaces it, since the pr was in a stale state and could not be merged

PR #158, opened 2017-01-30
Update GeoAxis middleware find correct header value

arhote:
Previously GeoAxis middleware used to look for OAM_REMOTE_USER
however the value that waitress gives to django is listed under
HTTP_OAM_REMOTE_USER. This change is necessary in order to
for Webgate to properly function and allow outside authentication

This is related to https://github.com/boundlessgeo/exchange/pull/156


davisc:
@arhote is this change necessary given `GEOAXIS_HEADER` is a configurable setting?

arhote:
@davisc I suppose given that it's configurable its not strictly necessary, but it makes sense to have the default be the correct value rather than having it be one more thing to worry about come config time.

BerryDaniel:
@arhote The correct value should be one without `HTTP_`, the remote header should not include that and if it does then the apache webgate configuration is not correct. If someone still wants to use `HTTP_` then it is configurable. Since it is configurable I am going to close this PR.

PR #171, opened 2017-02-12
Added fileservice, made sure attached images to features work in maploom. 

smesdaghi:
added fileservice to exchange and disabled streaming support with option to enable it. Updated settings.sh and bootstrap.sh to handle new vm creations properly which still need to be handled for rpm installations as well. Tested in maploom to make sure images attached to features now work.

### DIFF
--- a/docker/travis/Dockerfile.ci
+++ b/docker/travis/Dockerfile.ci
@@ -8,6 +8,9 @@ RUN chmod -Rf 755 /tmp/geoserver
 
 # install geonode-exchange and dependencies
 ADD . /opt/boundless/exchange
+RUN mkdir -p /opt/boundless/exchange/.storage/media/fileservice
+RUN chmod -Rf 777 /opt/boundless/exchange/.storage/media/fileservice
+
 RUN echo "GEOS_LIBRARY_PATH = '/opt/boundless/vendor/lib/libgeos_c.so'" >> /opt/boundless/exchange/exchange/settings/default.py
 
 # install python dependencies
--- a/exchange/fileservice/helpers.py
+++ b/exchange/fileservice/helpers.py
@@ -1,0 +1,42 @@
+from django.conf import settings
+import os
+
+def get_streaming_supported():
+    """
+    example settings file
+    FILESERVICE_CONFIG = {
+        'streaming_supported': True
+    }
+    """
+    conf = getattr(settings, 'FILESERVICE_CONFIG', {})
+    return conf.get('streaming_supported', False)
+
+
+def get_fileservice_dir():
+    """
+    example settings file
+    FILESERVICE_CONFIG = {
+        'store_dir': '/var/lib/geoserver_data/fileservice_store'
+    }
+    """
+    conf = getattr(settings, 'FILESERVICE_CONFIG', {})
+    dir = conf.get('store_dir', './fileservice_store')
+    return os.path.normpath(dir) + os.sep
+
+
+def get_fileservice_whitelist():
+    conf = getattr(settings, 'FILESERVICE_CONFIG', {})
+    return [x.lower() for x in conf.get('types_allowed', [])]
+
+
+def u_to_str(string):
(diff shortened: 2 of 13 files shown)


BerryDaniel:
@smesdaghi can you add tests?

docker/travis/Dockerfile.ci
```diff
@@ -8,6 +8,9 @@ RUN chmod -Rf 755 /tmp/geoserver
 
 # install geonode-exchange and dependencies
 ADD . /opt/boundless/exchange
+RUN mkdir -p /opt/boundless/exchange/.storage/media/fileservice
+RUN chmod -Rf 777 /opt/boundless/exchange/.storage/media/fileservice
```
BerryDaniel: Is 777 required? The process should only need to read the files in this directory.

exchange/fileservice/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_fileservice_dir():
+    """
+    example settings file
+    FILESERVICE_CONFIG = {
+        'store_dir': '/var/lib/geoserver_data/fileservice_store'
+    }
+    """
+    conf = getattr(settings, 'FILESERVICE_CONFIG', {})
+    dir = conf.get('store_dir', './fileservice_store')
```
BerryDaniel: we should set the default store to `os.path.join(settings.MEDIA_ROOT, 'fileservice')`

exchange/settings/default.py
```diff
@@ -393,6 +394,11 @@ def isValid(v):
         'data/'
     )
 
+FILESERVICE_CONFIG = {
+    'store_dir': os.getenv('FILESERVICE_MEDIA_ROOT', None),
```
BerryDaniel: Why set this to None, when in the helper it is set to `'./fileservice_store'`. If this is a feature that is going to be configured by default we should just set to `os.path.join(settings.MEDIA_ROOT, 'fileservice')`

exchange/tests/tasks_test.py
```diff
@@ -4,8 +4,9 @@
 import pytest
 from django.conf import settings
 
-@pytest.mark.skipif(settings.REGISTRYURL is None,
```
BerryDaniel: @bitner may still need this I wouldnt remove it

exchange/urls.py
```diff
@@ -26,11 +26,15 @@
 from geonode.urls import urlpatterns as geonode_urls
 from . import views
 from django.views.defaults import page_not_found
+from fileservice.api import FileItemResource
```
BerryDaniel: couldnt this be isolated by creating a urls.py in the fileservice app, rather than having this in the exchange urls.py?

exchange/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+from django import setup
```
BerryDaniel: why are you doing this? django.setup() is already done before any test are run?

smesdaghi: no, it is not done already and this was needed.

exchange/urls.py
smesdaghi: done

exchange/tests/tasks_test.py
smesdaghi: that line was removed by bitner when i was working with him so I'll leave out and if he wants it back we can add it.

exchange/fileservice/helpers.py
smesdaghi: done

exchange/settings/default.py
smesdaghi: done

docker/travis/Dockerfile.ci
smesdaghi: done. since test wasn't passing on container, we tried it. but the tests no longer need to actually hit the file system anyway.

smesdaghi:
a new pull request was created and merged.

PR #172, opened 2017-02-13
fix to category filters

bitner:
Fixes issue with category filters in unified search. For some reason Elastic Search is truncating categories to 13 characters, so this truncates the query to match. Should still figure out what is causing this.


bitner:
use categories_exact which is not tokenized by elastic search for aggregation and searching

PR #189, opened 2017-03-01
Added Active Directory backend

BerryDaniel:
- migrated ldap config to a new settings file named _ldap
- added ability to support ldap and ad
- added the ability to map group searches to django roles is_active, is_staff, and is_superuser
- adjusted str2bool to handle None values
- bumped version of geonode to include changes to permissions

### DIFF
--- a/exchange/settings/_ldap.py
+++ b/exchange/settings/_ldap.py
@@ -36,13 +36,12 @@ if str2bool(os.environ.get('AUTH_LDAP_DEBUG')):
 
 AUTH_LDAP_SERVER_URI = os.environ.get('AUTH_LDAP_SERVER_URI', None)
 LDAP_SEARCH_DN = os.environ.get('LDAP_SEARCH_DN', None)
+LDAP_EMAIL_MAP = os.environ.get('LDAP_EMAIL_MAP', 'mail')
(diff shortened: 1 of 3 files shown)


exchange/settings/_ldap.py
```diff
@@ -0,0 +1,100 @@
...
+
+    logger = logging.getLogger('django_auth_ldap')
+    logger.addHandler(logging.StreamHandler())
+    logger.setLevel(logging.DEBUG)
+
+AUTH_LDAP_SERVER_URI = os.environ.get('AUTH_LDAP_SERVER_URI', None)
+LDAP_SEARCH_DN = os.environ.get('LDAP_SEARCH_DN', None)
+if str2bool(os.environ.get('LDAP_IS_AD', 'False')):
+    AUTH_LDAP_USER = '(SAMAccountName=%(user)s)'
+    LDAP_EMAIL_MAP = 'userPrincipalName'
```
amirahav: It might be a good idea to make ```LDAP_EMAIL_MAP``` an environment variable. 
In most cases  ```userPrincipalName``` is the internal email address, and other deployments may want to use the ```mail``` attribute instead

BerryDaniel: Made the adjustment let me know your thoughts.

PR #190, opened 2017-03-02
Fixed incorrect empty search text messages for maps/layers

travislbrundage:
This is a little ugly but because this template is reused in multiple places and the dataValue variable is reused in those places to mean completely different things, it's placing an ad hoc if check to see if we should use dataValue as part of the message to display or not.

Alternatively, can just use the word "content" or something to have a generic static message instead.


davisc:
@travislbrundage to simply, lets go with your second option of generalizing the message to 'No content create'

travislbrundage:
Updated.

PR #201, opened 2017-03-07
codecov.io integration

BerryDaniel:
adjusted coverage reporting to use codecov.io not coveralls.io, codecov has a bash command that allows publishing the report within the container. Updated the badge in the readme.


BerryDaniel:
@bitner @amirahav @davisc disregard the codecov error, it is showing red, because it is the first time and it doesnt have anything to compare with.

PR #203, opened 2017-03-08
Allow setting of GEOAXIS_HEADER via ENV VARS

arhote:
The value that Geoaxis returns as its header for
authentication is customizable and can vary from
instance to instance. As a result the user of
Exchange should be able to customize the value
Exchange looks for to match their flavor of
Geoaxis.


BerryDaniel:
Thanks @arhote

PR #207, opened 2017-03-15
"Workspace" addition bugs

theduckylittle:
In the new search screen it was possible to add registry layers and maps to the "Workspace"/cart.  These things are broken functionality in the current application.  This PR prevents the user from performing those actions. A bit of detail:

1. Fixes from CSS with the cart items when the title gets too long.
2. Filter out "add" buttons from items which are not layers.
3. Ensure all filters and the "Create a Map" button can be displayed at the same time.

Depends on : 
- https://github.com/boundlessgeo/geonode/pull/39

Screenshot:
image

### DIFF
--- a/exchange/static/css/site_base.css
+++ b/exchange/static/css/site_base.css
@@ -143,3 +143,21 @@ input[value='Invite User'] {
 span[id^="error_"].help-inline {
   color: #d9534f;
 }
+
+resource-cart span.cart-label {
+    display: inline-block;
+    overflow: hidden;
+}
+
+
+@media screen and (min-width: 960px) {
+    resource-cart span.cart-label {
+        max-width: 150px;
+    }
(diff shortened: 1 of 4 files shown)


davisc:
@theduckylittle looks like this PR prevents registry results from being added to the "workspace" and from the 'create a map' functionality. We need to support both of those use cases per client request.

theduckylittle:
@davisc Yes, it does prevent that.  The bug as filed (NODE-735) was that there is broken functionality.  I was hoping to get the application fixed before adding a new feature (adding registry layers to a map directly from search) in a separate PR after the basics were working.

So, if we can merge this PR and I can open a new JIRA story that can handle the addition of registry layers from search.

davisc:
Sounds like a plan.

PR #213, opened 2017-03-20
Added additional CSS to prevent fly-by message.

theduckylittle:
Inline CSS does not need to wait for CSS or Angular to load to
hide the contents of the feature datasets.  Only once the layer
list has loaded will the list display.


theduckylittle:
Refs: NODE-804

PR #222, opened 2017-03-25
Consent Banner for Exchange

davisc:
This PR does the following:

- Adds a consent page to the login workflow
- Consent page text is configurable
- Consent page supports two authentication providers 1) Exchange local and 2) GeoAxis
- Hides header nav when LOCKDOWN is enabled and displays it once the user has authenticated
- Switched text Sign in to Log In to be consistent with the rest of the application

**EXCHANGE_LOCKDOWN mode enabled and user not authenticated**

screen shot 2017-03-25 at 1 17 30 pm

**EXCHANGE_LOCKDOWN enabled, LOGIN_WARNING enabled, GEOAXIS enabled and user not authenticated**

screen shot 2017-03-25 at 12 54 42 pm

**EXCHANGE_LOCKDOWN enabled, LOGIN_WARNING enabled, GEOAXIS DISABLED and user not authenticated** Defaults to username password authentication

screen shot 2017-03-25 at 1 07 50 pm

screen shot 2017-03-25 at 1 10 24 pm

### DIFF
--- a/exchange/settings/default.py
+++ b/exchange/settings/default.py
@@ -474,3 +474,25 @@ if ENABLE_SOCIAL_LOGIN:
         AUTHENTICATION_BACKENDS += (
             'exchange.auth.backends.geoaxis.GeoAxisOAuth2',
         )
+        
+        LOGIN_WARNING = True
+        LOGIN_WARNING_TEXT = '<p>You are accessing a U.S. Government (USG) Information System (IS) that is provided ' \
+                     'for USG-authorized use only.  By using this IS ' \
+                     '(which includes any device attached to this IS), you ' \
+                     'consent to the following conditions:</p><ul><li>The USG routinely intercepts, and monitors ' \
+                     'communications on this IS for purposes including, but not limited to, penetration testing, ' \
+                     'COMSEC monitoring, network operations and defense, personnel misconduct (PM), law enforcement ' \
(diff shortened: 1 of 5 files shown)


exchange/settings/default.py
```diff
@@ -474,3 +474,20 @@ def isValid(v):
         AUTHENTICATION_BACKENDS += (
             'exchange.auth.backends.geoaxis.GeoAxisOAuth2',
         )
+        
+        LOGIN_WARNING = True
```
BerryDaniel: We should pull this via os.environ and set this as false by default.

exchange/settings/default.py
```diff
@@ -474,3 +474,20 @@ def isValid(v):
...
+        LOGIN_WARNING_TEXT = '''<p>You are accessing a U.S. Government (USG) Information System (IS) that is provided
```
BerryDaniel: We should also set this as an os.environ too, but have this as the default. It would be good to have a check `if LOGIN_WARNING:` then enable this setting. With this current setup everyone would habe the consent banner.

exchange/settings/default.py
davisc: Will do.

BerryDaniel:
image

The consent banner in chrome, we may need to make some adjustments to the css to support other browsers. thoughts? @davisc

PR #225, opened 2017-03-27
Make info tab into a proper tab to be consistent with other UI elements

travislbrundage:
Addresses [NODE-794](https://issues.boundlessgeo.com:8443/browse/NODE-794)


walker:
@travislbrundage: These tabs are great!

PR #234, opened 2017-04-04
Revamped Celery/Task tests

theduckylittle:
1. Created new 'celery_test.py' which tests celery and the
   CSW task from exchange.
2. Removed the old (borken) tasks_test.py


BerryDaniel:
@theduckylittle I think we need to adjust travis to ensure that an amqp service is running.

theduckylittle:
@BerryDaniel I've updated these tests to check whether unified search is enabled or not. This has the net effect of dropping the coverage % but the related tests pass cleanly.  The tests do run successfully in a dev environment.  We will need to revise the docker environment(s) to include registry properly if we intend for this test to be used in travis.

PR #235, opened 2017-04-05
use docs from github master

BerryDaniel:
@arhote


arhote:
@BerryDaniel 

1. Travis isn't happy with something you did.
2. Are we tagging version numbers in the docs repo so that we can point to something other than just master?

BerryDaniel:
@arhote looks like it just chocked on the install, restarted the job and it is passing now

PR #239, opened 2017-04-11
Renderer Switching

j-meyer:
Allows users to switch the renderer of a map rather than having it as a server setting.

screen shot 2017-04-11 at 10 05 39 am


walker:
@davisc Should we consider this as a component of bridging to SDK as default?

davisc:
@walker Yes. Based on some of my more recent conversations, I can see the requirement evolving from here but this functionality supports SDK 1.0 client, MapLoom, Etc.

davisc:
Closing this PR as it has been pulled into another branch.

PR #242, opened 2017-04-11
The Promised Thumbnail API

theduckylittle:
A new (simplified) Thumbnail API for Exchange!

1. There are tests for all of the Python code.  The coverage is nearly 100% with only a few edge cases that are difficult to simulate in the test environment.
2. The API is truly pretty simple, there is one end point. /thumbnails/[object_type]/[object_id].  E.G. /thumbnails/layers/geonode:transit_stops
  a. GET from that end point returns an image. A blank GIF if not set.
  b. POST takes the contents of an image file and assigns that as the thumbnail to the specified type/id.
3. There is a new Thumbnail model which stores the image in the database.
4. There is a new signal post-save hook for maps and layers to update their images when they are saved.  
5. If a user sets a legend then that legend *will not* be overwritten by the automatic legend generator.

## Screenshots

image

image

Example of a thumbnail uploaded from a file...
image

And that thumbnail showing up in the listing....
image


theduckylittle:
Another screen shot showing how the thumbnail uploader was removed from the metadata details screen:
image

bitner:
This is a bridge commit, test failures were caused by another merge will be fixed in a separate branch.

theduckylittle:
PR to fix the broken tests is here:
- https://github.com/boundlessgeo/exchange/pull/243

PR #251, opened 2017-04-18
added nearsight configuration

jsvrcek:
The configuration provided is an example configuration of NearSight inside of Exchange.  
Adding `NEARSIGHT` to the environment will add the nearsight URLs.
Then migrations will need to be ran.  
Lastly in order to work properly a directory needs to be provided for `NEARSIGHT_UPLOAD` with permissions given to the django server to read and write to the directory.


dev/settings.sh
```diff
@@ -28,4 +28,7 @@ export SOCIAL_BUTTONS='False'
 export SECRET_KEY='exchange@q(6+mnr&=jb@z#)e_cix10b497vzaav61=de5@m3ewcj9%ctc'
 export DEFAULT_ANONYMOUS_VIEW_PERMISSION='False'
 export DEFAULT_ANONYMOUS_DOWNLOAD_PERMISSION='False'
+export NEARSIGHT=True
```
smesdaghi: change to NEARSIGHT_ENABLED

PR #274, opened 2017-05-03
Adds osgeo importer handler to add time extents

bitner:


exchange/importer/geonode_timeextent_handler.py
```diff
@@ -0,0 +1,84 @@
...
+from osgeo_importer.handlers import ImportHandlerMixin
+from osgeo_importer.handlers import ensure_can_run
+from osgeo_importer.models import UploadLayer
+from osgeo_importer.utils import quote_ident
+from geonode.layers.models import Layer
+from osgeo_importer.handlers.geonode.backward_compatibility import set_attributes
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+logger = logging.getLogger('rootqq')
```
davisc: @bitner which logger is `rootqq`?

exchange/importer/geonode_timeextent_handler.py
```diff
@@ -0,0 +1,84 @@
...
+
+        conn = db.connections[settings.OSGEO_DATASTORE]
+        cursor = conn.cursor()
+        scrub_layer_name = quote_ident(layer)
+        logger.debug('Getting min/max for %s', scrub_layer_name)
+
+        mint = None
+        maxt = None
+
+        if self.has_start and self.has_end:
```
davisc: This would be useful in the updateLayers management command. I will log an issue.

exchange/importer/geonode_timeextent_handler.py
bitner: d'oh, was wrestling getting logging to work, will change to __name__

PR #278, opened 2017-05-03
Node 884 - Improve CSW form

arhote:
- The first entry in the CSW Insert form is for the source field
- The "Source" entry is relabeled "Service Endpoint URL"
- The "Creator" entry is replaced with "Agency"
- Remove "Contact" entry
- Remove "Relation" entry
- Add "Email" entry
- Add "Phone" entry

### DIFF
--- a/exchange/core/models.py
+++ b/exchange/core/models.py
@@ -100,7 +100,7 @@ class CSWRecord(models.Model):
     record_type = models.CharField(max_length=128, blank=True)
     alternative = models.CharField(max_length=128, blank=True)
     abstract = models.TextField(blank=True)
-    source = models.URLField(max_length=512, blank=False)
+    service_endpoint_url = models.URLField(max_length=128, blank=False)
     relation = models.CharField(max_length=128, blank=True)
     record_format = models.CharField(max_length=128, blank=True)
     bbox_upper_corner = models.CharField(max_length=128,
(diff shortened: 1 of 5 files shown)


exchange/core/models.py
```diff
@@ -95,21 +95,21 @@ class CSWRecord(models.Model):
     classification = models.CharField(max_length=128, blank=True)
     title = models.CharField(max_length=328, blank=False)
     modified = models.DateField(default=datetime.date.today, blank=False)
-    # 'creator' is assumed to be distinct from logged-in User here
-    creator = models.CharField(max_length=328, blank=True)
+    # 'agency' is assumed to be distinct from logged-in User here
```
davisc: Changing the field will require a migration entry so that preexisting fields can be migrated. Instead, since this change set is just cosmetic, we can use 'Field.verbose_name' to reflect the label change. For example, `creator = models.CharField(max_length=328, blank=True, verbose_name='Agency'). We can follow the same approach for source, relation, contact_information, etc.

exchange/core/models.py
```diff
@@ -95,21 +95,21 @@ class CSWRecord(models.Model):
...
+    agency = models.CharField(max_length=328, blank=True)
     record_type = models.CharField(max_length=128, blank=True)
     alternative = models.CharField(max_length=128, blank=True)
     abstract = models.TextField(blank=True)
-    source = models.URLField(max_length=512, blank=False)
```
davisc: Since source is pull from the csw standard, we can leave it as is and just make the label reflect the text 'Source URL'

exchange/core/models.py
```diff
@@ -95,21 +95,21 @@ class CSWRecord(models.Model):
...
-    relation = models.CharField(max_length=128, blank=True)
+    service_endpoint_url = models.URLField(max_length=512, blank=False)
     record_format = models.CharField(max_length=128, blank=True)
     bbox_upper_corner = models.CharField(max_length=128,
                                          default="85.0 180",
                                          blank=True)
     bbox_lower_corner = models.CharField(max_length=128,
                                          default="-85.0 -180",
                                          blank=True)
-    contact_information = models.CharField(max_length=128, blank=True)
```
davisc: Removing this field may cause the celery task to fail. Need to check the csw schema for a mapping on contact_email and contact_phone if we are breaking it out into distinct fields.

theduckylittle:
1. The migrations look fine and executed cleanly on my local database.
2. The celery task is passing and using @property appears to be working as expected for populating contact_information.

PR #305, opened 2017-06-08
Create LICENSE file

bartvde:
Add a LICENSE file to make it more obvious what the license is (https://github.com/boundlessgeo/exchange/blob/master/setup.py#L47)


BerryDaniel:
Nice catch @travislbrundage and thanks for the quick turn around @bartvde

PR #306, opened 2017-06-10
Added about tab with info about Exchange and dependent projects

travislbrundage:
Includes a new About tab for Boundless Exchange. The tab is accessible from the footer similarly to the Documentation page. Additionally, it is accessible in the "About" dropdown menu in the header.

The new view grabs and displays information about the community and sub projects Exchange relies on, including version numbers, contributors from Boundless, the commit hash being used, and so on. The about tab is designed to show what is new with the version of Exchange and the current versions of the dependent projects. Images below.

image
image
image

**NOTE**: The following PR's are required for the Version Dependency section to show versions and commits correctly.

https://github.com/boundlessgeo/geonode/pull/62 [MERGED]
https://github.com/boundlessgeo/django-exchange-maploom/pull/6 [OPEN]
https://github.com/GeoNode/django-osgeo-importer/pull/159 [OPEN]
https://github.com/GeoNode/geonode-client/pull/174 [MERGED]

### DIFF
--- a/exchange/themes/templates/about.html
+++ b/exchange/themes/templates/about.html
@@ -1,0 +1,35 @@
+{% extends "site_base.html" %}
+{% load i18n %}
+
+{% block title %} {% trans "About" %} - {{ block.super }} {% endblock %}
+
+{% block middle %}
+{% block mainbody %}
+<div class="container" style="text-align: center">
+  <h1>Boundless Exchange</h1>
+  v{{ exchange_version }}
+  <br>
+  <p>Boundless Exchange is a web-based platform for your content, built for your enterprise. It facilitates the creation, sharing, and collaborative use of geospatial data. For power users, advanced editing capabilities for versioned workflows via the web browser are included. Boundless Exchange is powered by GeoNode, GeoGig, OpenLayers, PostGIS and GeoServer. Boundless Exchange is designed as a platform for collaboration.</p>
+</div>
+<div class="container">
+  <h2>What's New</h2>
+  <!-- TODO: only display link if the Exchange instance can reach open Internet -->
+  <h4>Boundless Exchange -- <a href="https://github.com/boundlessgeo/exchange/releases/tag/v{{ exchange_version }}">{{ exchange_version }}</h4></a>
+  <p>{{ exchange_release|linebreaks }}</p>
+</div>
+<div class="container">
+  <h2>Version Details</h2>
+  <ul>
+    {% for project in projects %}
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} version: v{{ project.version }}</h4></li></div>
+    <!-- TODO: only display link if the Exchange instance can reach open Internet -->
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} code revision: {% if project.boundless_repo %}<a href="{{ project.boundless_repo }}/commit/{{ project.commit }}">{% else %}<a href="{{ project.repo }}/commit/{{ project.commit }}">{% endif %}{{ project.commit }}</a></h4></li></div>
+    {% endfor %}
+  </ul>
+</div>
+<div class="container">
+  <h2>License</h2>
+  <p>Boundless Exchange is free software licensed under GNU General Public License (version 3). Refer to the included license.txt file for the full license.</p>
+</div>
+{% endblock %}
+{% endblock %}
--- a/exchange/views.py
+++ b/exchange/views.py
@@ -23,6 +23,9 @@ from django.views.generic import CreateView, UpdateView, ListView
 
 from django.core.urlresolvers import reverse_lazy
 from django.db import transaction
+from pip._vendor import pkg_resources
+from exchange.version import get_version
+from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


arhote:
@travislbrundage Looks good to me, other than the need to rebase to master I'm down to merge. @davisc any objections?

exchange/themes/templates/about.html
```diff
@@ -0,0 +1,63 @@
...
+{% block mainbody %}
+<div class="container" style="text-align: center">
+  <h1>Boundless Exchange</h1>
+  v{{ exchange_version }}
+  <br>
+  <p>Boundless Exchange is a web-based platform for your content, built for your enterprise. It facilitates the creation, sharing, and collaborative use of geospatial data. For power users, advanced editing capabilities for versioned workflows via the web browser are included. Boundless Exchange is powered by GeoNode, GeoGig, OpenLayers, PostGIS and GeoServer. Boundless Exchange is designed as a platform for collaboration.</p>
+</div>
+<div class="container">
+  <h2>What's New</h2>
+  <h4>Boundless Exchange -- <a href="https://github.com/boundlessgeo/exchange/releases/tag/v{{ exchange_version }}">{{ exchange_version }}</h4></a>
```
travislbrundage: Should this href be wrapped in some condition for systems which cannot reach out to the open Internet?

exchange/themes/templates/about.html
```diff
@@ -0,0 +1,63 @@
...
+    {% endif %}
+  </div>
+  {% endfor %}
+</div>
+<div class="container">
+  <h2>Version Details</h2>
+  <ul>
+    {% for project in projects %}
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} version: v{{ project.version }}</h4></li></div>
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} code revision: {% if project.boundless_repo %}<a href="{{ project.boundless_repo }}/commit/{{ project.commit }}">{% else %}<a href="{{ project.repo }}/commit/{{ project.commit }}">{% endif %}{{ project.commit }}</a></h4></li></div>
```
travislbrundage: Echo above question

exchange/views.py
```diff
@@ -36,6 +38,92 @@ def documentation_page(request):
     return HttpResponseRedirect('/static/docs/index.html')
 
 
+def about_page(request, template='about.html'):
+    exchange_version = get_version()
```
amirahav: For exchange version default to ```geonode-exchange``` from pip, and fall back to ```get_version()```

travislbrundage: Okay - will update that.

exchange/views.py
```diff
@@ -36,6 +38,92 @@ def documentation_page(request):
...
+            exchange_release = release['body'].replace(' - ', '\n-')
+
+    geonode_version = [
+        p.version for p in pkg_resources.working_set
+        if p.project_name == 'GeoNode'
+    ][0]
+
+    # Correct way to request to local?
+    geoserver_version = requests.get(
+        'http://172.16.238.2/geoserver/rest/about/version.json'
```
amirahav: use OGC_SERVER['default']['LOCATION'] for geoserver url

travislbrundage: This is my only open question - looking into it myself and will update if I cannot figure it out.

travislbrundage: Thank you! Updating.

exchange/views.py
```diff
@@ -23,6 +23,9 @@
 
 from django.core.urlresolvers import reverse_lazy
 from django.db import transaction
+from pip._vendor import pkg_resources
+from exchange.version import get_version
+from exchange.settings import OGC_SERVER
```
amirahav: This should be ``` from django.conf import settings``` so if OGC_SERVER gets overridden somewhere else it still works

exchange/views.py
```diff
@@ -36,6 +39,100 @@ def documentation_page(request):
...
+    for release in exchange_releases:
+        if release['tag_name'] == 'v' + exchange_version[:-8]:
+            exchange_release = release['body'].replace(' - ', '\n-')
+
+    geonode_version = [
+        p.version for p in pkg_resources.working_set
+        if p.project_name == 'GeoNode'
+    ][0]
+
+    geoserver_version = requests.get(
```
amirahav: This should be wrapped with a try block, and should probably return an empty string on failure

exchange/views.py
```diff
@@ -36,6 +39,100 @@ def documentation_page(request):
...
+        settings.OGC_SERVER['default']['LOCATION'] + 'rest/about/version.json'
+    ).json()['about']['resource'][0]
+    geoserver_commit = geoserver_version['Git-Revision'][:7]
+    geoserver_version = geoserver_version['Version'] + '.' + geoserver_commit
```
amirahav: pet peeve - use string formatting (```str.format```) instead of string concatenation. (And anywhere else in this PR this exist)

exchange/themes/templates/about.html
amirahav: yes, this should probably wrapped with something for that case

exchange/themes/templates/about.html
amirahav: yes, this should probably wrapped with something for that case

exchange/views.py
```diff
@@ -36,6 +39,100 @@ def documentation_page(request):
...
+        exchange_version = get_version()
+    else:
+        exchange_version = exchange_version[0]
+    try:
+        exchange_releases = requests.get(
+            'https://api.github.com/repos/boundlessgeo/exchange/releases'
+        ).json()
+    except:
+        exchange_releases = []
+    exchange_release = 'No release notes available.'
```
amirahav: Having both plural and singular (```exchange_release``` and ```exchange_releases```) is confusing.
Maybe change ```exchange_release``` to ```release_notes```?

amirahav:
I'm -1 on the community projects page. I think that it just clutters the page, and confuses users without adding any value to the users.
This information should be on our website instead.

exchange/views.py
travislbrundage: Good point. Updated to call it `release_notes` instead.

exchange/views.py
travislbrundage: Sure thing - will adjust string concatenation to use string formatting.

exchange/views.py
travislbrundage: Done, thanks for the suggestion!

travislbrundage: Good point, I forgot when doing this GET. Fixed.

exchange/views.py
travislbrundage: Updated these.

travislbrundage:
Yeah, I think it does not make sense to have it anymore (originally was using it to show contributors as well, but with that scrapped, little point). Looking again, it looks fine without it. I will update screenshots when I rebase & push the changes you mentioned.

Thanks for your attention to detail and reviewing my code, Ami!

exchange/themes/templates/about.html
```diff
@@ -0,0 +1,35 @@
...
+<div class="container" style="text-align: center">
+  <h1>Boundless Exchange</h1>
+  v{{ exchange_version }}
+  <br>
+  <p>Boundless Exchange is a web-based platform for your content, built for your enterprise. It facilitates the creation, sharing, and collaborative use of geospatial data. For power users, advanced editing capabilities for versioned workflows via the web browser are included. Boundless Exchange is powered by GeoNode, GeoGig, OpenLayers, PostGIS and GeoServer. Boundless Exchange is designed as a platform for collaboration.</p>
+</div>
+<div class="container">
+  <h2>What's New</h2>
+  <!-- TODO: only display link if the Exchange instance can reach open Internet -->
+  <h4>Boundless Exchange -- {% if True %}<a href="https://github.com/boundlessgeo/exchange/releases/tag/v{{ exchange_version }}">{{ exchange_version }}</h4></a>{% else %}{{ exchange_version }}</h4>{% endif %}
```
travislbrundage: Only thing left is to fix these - instead of `if True`, want to display only if the Exchange instance can reach open Internet.

exchange/themes/templates/about.html
```diff
@@ -0,0 +1,35 @@
...
+  <p>{{ exchange_release|linebreaks }}</p>
+</div>
+<div class="container">
+  <h2>Version Details</h2>
+  <ul>
+    {% for project in projects %}
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} version: v{{ project.version }}</h4></li></div>
+    <!-- TODO: only display link if the Exchange instance can reach open Internet -->
+    <div style="float: left; width: 50%"><li><h4>{{ project.name }} code revision: {% if True %}{% if project.boundless_repo %}<a href="{{ project.boundless_repo }}/commit/{{ project.commit }}">{% else %}<a href="{{ project.repo }}/commit/{{ project.commit }}">{% endif %}{{ project.commit }}</a>{% else %}{{ project.commit }}{% endif %}</h4></li></div>
```
travislbrundage: Only thing left is to fix these - instead of `if True`, want to display only if the Exchange instance can reach open Internet.

exchange/views.py
```diff
@@ -36,6 +39,106 @@ def documentation_page(request):
...
+
+    try:
+        geoserver_version = requests.get(
+            '{}rest/about/version.json'.format(
+                settings.OGC_SERVER['default']['LOCATION']
+            )
+        ).json()['about']['resource'][0]
+    except:
+        geoserver_version = ''
+    geoserver_version = '{}.{}'.format(
```
amirahav: Sorry, but a few more :-(
This needs to be moved into the try statement so it won't raise an error when it's an empty string

exchange/views.py
```diff
@@ -36,6 +39,106 @@ def documentation_page(request):
     return HttpResponseRedirect('/static/docs/index.html')
 
 
+def about_page(request, template='about.html'):
+    exchange_version = [
+        p.version for p in pkg_resources.working_set
```
amirahav: How about moving these (getting the pip version of each package) into their own method? Might be a bit cleaner

exchange/views.py
travislbrundage: Good catch. Fixed.

exchange/views.py
travislbrundage: Sure thing. That does look a bit cleaner in the about_page function. Adjusted this.

PR #311, opened 2017-06-22
Add place to include tracking scripts when desired

arhote:
For certain deploys, it is desireable to include
scripts for tracking and analytics to the Exchange
site. This commit adds tracking_header.html and
tracking_footer.html as the defacto place to put
that scripting.


amirahav:
Do we only want the tracking header/footer in ```base.html``` ?

arhote:
@amirahav Doesn't everything extend base? Isn't that why it's called base?

amirahav:
image

PR #316, opened 2017-07-06
Registry category

bitner:
This commit uses the recent commit to Registry (https://github.com/boundlessgeo/registry/pull/146) to allow for adding custom mappings to Registry when registry is initialized. Adding or changing these mappings requires rebuilding the Registry Elastic Search instance.

To enable there is a setting "REGISTRY_MAPPINGS_OVERRIDES_FILE" that needs to point to a json file containing the elastic search mappings to be merged with the default Registry Mappings.

This PR fixes https://issues.boundlessgeo.com:8443/browse/NODE-970 by adding a copy of the Registry category using the same field as in the Haystack Elastic Search.

This PR fixes https://issues.boundlessgeo.com:8443/browse/NODE-881 by adding a source type filter and facet. In order to show up in the drop down, additional options are added by hand to the template as the existing template relies on calls to just the geonode backend to fill the options on initial page load not allowing for adding new types through the use of the facets json returned by the search API.

Additionally, this PR simplifies the query code by being able to use a single field name for date which also allows for being able to sort by date across both geonode and registry items (rather than geonode first by date followed by registry by date)


walker:
> Looks good. Some changes will be required within Chef recipes.

@camerontesterman: Can you create the sub-task(s) for this in JIRA with regard to these modifications?

camerontesterman:
@walker working on these now.

PR #329, opened 2017-07-21
Fix issues with alerts/warnings being hidden

arhote:
Recently user notifications were changed to be a
new alert class. This change ensures that when the
classification banner is enabled that the alert
class is shifted down enough that it's not hidden
by the nav bar.


walker:
@arhote: Can we get a screenshot?

arhote:
@walker Sure!

Before:
screen shot 2017-07-21 at 11 01 13 am

After:
screen shot 2017-07-21 at 11 01 58 am

walker:
Thanks, @arhote!

PR #331, opened 2017-07-24
Make release notes expandable/collapsible

travislbrundage:
Addressing feedback: Adds a button to the about page under "What's New?" which will expand to show release notes (prior, all release notes were showing by default, which were sometimes long and stretched the page).


davisc:
@travislbrundage Can you add a screenshot?

PR #333, opened 2017-07-26
Adjust promise resolution with bump to angular 1.6.5

travislbrundage:
Only remaining error that I can see:
image

Promise handling seems to be the only change which needed to be addressed.

Coincides with: https://github.com/boundlessgeo/geonode/pull/70

### DIFF
--- a/exchange/static/bower.json
+++ b/exchange/static/bower.json
@@ -3,7 +3,6 @@
   "version": "1.1.0",
   "license": "GPLv3+",
   "dependencies": {
-    "clipboard": "~1.5.12",
-    "angular-cookies": "1.4.0"
+    "clipboard": "~1.5.12"
   }
 }
--- a/exchange/static/osgeo_importer/importer.js
+++ b/exchange/static/osgeo_importer/importer.js
@@ -90,15 +90,15 @@
     this.configureUpload = function(layer) {
       var deferredResponse = q_.defer();
       this.validateConfigurationOptions(layer);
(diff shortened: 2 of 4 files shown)


exchange/static/bower.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.1.0",
   "license": "GPLv3+",
   "dependencies": {
-    "clipboard": "~1.5.12",
-    "angular-cookies": "1.4.0"
```
travislbrundage: This is handled by geonode and is no longer pointing to the correct version

exchange/static/osgeo_importer/importer.js
```diff
@@ -90,15 +90,15 @@
...
         deferredResponse.resolve(layerService_.update(layer));
-      }).error(function(data, status) {
+      }, function(response) {
           var error = 'Error configuring layer.';
           if (data.hasOwnProperty('error_message')) {
-              error = data.error_message;
+              error = response.data.error_message;
           }
-          deferredResponse.reject(error, data, status);
+          deferredResponse.reject(error, response.data, response.status);
```
travislbrundage: Does this reject call make sense? Even in angular 1.3.0 from the docs I only see reject taking one argument, so I'm not sure I understand what's happening here

walker: @travislbrundage: afaik, deferred reject promise only takes the response object.

PR #335, opened 2017-07-27
Update quick copy link

davisc:
Link to WMS of uploaded GeoNode layer should copy specific reference


amirahav:
Hold off on this, both the WMS and WFS and not correct.
WMS should be geoserver/{workspace}/{layer_name}/ows?service=wms&version=1.3.0&request=GetCapabilities

arhote:
@amirahav Sorry, I didn't see your comment before I hit merge, I'll reopen and have @davisc Update.

davisc:
@amirahav after thinking about it, if the layer is not public, GS will return unknown feature type.

amirahav:
@davisc Right, but that would be the case for any layer specific request we make.
The layer must be public, or the request must include an access token.
Are you thinking we should just return the global getcaps instead of a layer specific one, just so the user doesn't see an error?

amirahav:
Another option, which really don't like, would be to always allow gecaps on all services and layers, regardless of how exchange configures security.

This should work with Geofence, however will end up being very confusing for users.

PR #336, opened 2017-07-27
Update exchange for airgapped installs to remove links to boundlessge…

rcrissup:
Added new variable so that installs on closed networks will have the URL to boundlessgeo.com disabled

### DIFF
--- a/exchange/core/context_processors.py
+++ b/exchange/core/context_processors.py
@@ -58,6 +58,7 @@ def resource_variables(request):
         GEOQUERY_ENABLED=getattr(settings, 'GEOQUERY_ENABLED', False),
         GEOQUERY_URL=getattr(settings, 'GEOQUERY_URL', None),
         EXTENT_FILTER_ENABLED=getattr(settings, 'EXTENT_FILTER_ENABLED', True),
+        ON_AIRGAPPED_NETWORK=getattr(settings, 'ON_AIRGAPPED_NETWORK', False),
     )
 
     return defaults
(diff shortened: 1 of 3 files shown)


amirahav:
My two suggestions of dealing with external links:

1. I think that in the case of a the link back to our homepage we could/should drop it instead of adding another config option to try on hide it.

2. If we don't want to go with the above, then adding a method to check internet connectivity and then hiding the link based on that would make more sense. 
We should generally attempt to have as little configuration options as possible to make it easier on our users.

walker:
> If we don't want to go with the above, then adding a method to check internet connectivity and then hiding the link based on that would make more sense.

This might be preferred based on below.

> We should generally attempt to have as little configuration options as possible to make it easier on our users.

I appreciate this thought and agree.

travislbrundage:
Just trying to generate a bit more literal direction here, should this be done in the view function and passed in through the context dict?

walker:
@amirahav ^ see Travis' question.

walker:
@rcrissup: based on discussion during standup:

1. Sticking with build/install configuration variable for this
2. Changing the configuration variable name to be more specific/self-descriptive
3. Looking to see documentation

davisc:
@travislbrundage its being added to the scope via the context processor, is that what you are referring to?

amirahav:
I think @travislbrundage question had more to do with my suggestion of creating a method to check network access.
But with @walker's comment above this is now pretty straight forward - replace ```ON_AIRGAPPED_NETWORK``` with something like ```SHOW_EXTERNAL_LINKS``` (or whatever) , and then see if there are other links that should be treated like this.

walker:
Sorry @rcrissup, when the PR popped up in my feed, I didn't realize @arhote had already provided feedback on the getattr items. Didn't mean to double-team.

arhote:
@davisc Any idea why this PR would be tanking the code coverage like that? He shouldn't have touched anything to make all those files go sideways.

PR #346, opened 2017-08-22
Added settings variable to inform geonode-client to request map tiles…

travislbrundage:
… cross origin

[NODE-1082](https://issues.boundlessgeo.com:8443/browse/NODE-1082)

Related PR's: https://github.com/boundlessgeo/sdk/pull/618


exchange/settings/default.py
```diff
@@ -595,3 +595,4 @@ def isValid(v):
     NOMINATIM_ENABLED = True
 
 SEARCH_FILTERS['HOST_ENABLED'] = True
+CROSS_ORIGIN = str2bool(os.getenv('CROSS_ORIGIN', 'False'))
```
travislbrundage: Should this just be directly assigned instead of using an environment variable?

travislbrundage: Going to change this to have the name `MAP_CLIENT_USE_CROSS_ORIGIN_CREDENTIALS` as has been requested on the other PR's.

PR #351, opened 2017-09-04
Search refactor

bitner:
This PR is a major refactor of the way that Search works, particularly with regard to Faceting. All facets are now controlled solely by the configuration in the View with a single data-bound view that goes through all facets. There are likely some differences in how some things work in this PR that will need additional tickets to address based on capabilities offered by this approach. Solely using Elastic Search to control facets, there is a limit on the number of facets that are returned. That limit is currently defaulted to 15, but can be controlled by the "nfacets" url parameter. This is currently not exposed in the UI and the expansion for keywords has been removed, but this parameter could be used to fulfill the same roll. With the default, the 15 options with the highest count will be returned and used for the facets.


davisc:
@bitner looks good, I will merge it so that we can get it deployed to dev.

PR #360, opened 2017-09-11
Update warning banner based on feedback

amirahav:


walker:
This is only for ps/nga branch, no?

davisc:
The banner text is customizable so it could be used for a variety of situations.

arhote:
Can we get before and after screenshots since this is mostly a graphical change?

PR #364, opened 2017-09-14
Limit the cart/worksapce cookie to only include required info

amirahav:


travislbrundage:
For some reason this won't remove the cookie anymore, so they are stuck in the workspace. I am not immediately sure why.

PR #370, opened 2017-09-20
Mac docker file

travislbrundage:
Updates to docker-compose.yml file to work natively on Mac OS X.

Tested and works on Ubuntu 16.04, so the solution should have cross OS compatibility.


bitner:
Hey Travis, We need to make sure that we include the registry setup in here.

travislbrundage:
Gonna close just real quick while I smoke test so it doesn't get badly merged. Updating based on your feedback, thanks bitner!

PR #374, opened 2017-09-21
Workaround for elasticsearch indexing - change importer to offer publ…

travislbrundage:
…ishing the layer

Coincides with a PR to default imported layers as unpublished, but does not require it.

https://github.com/GeoNode/django-osgeo-importer/pull/177

### DIFF
--- a/exchange/static/osgeo_importer/partials/uploadWizard.html
+++ b/exchange/static/osgeo_importer/partials/uploadWizard.html
@@ -163,11 +163,13 @@
                     <div class="step-title" ng-hide="success"><h4>Ready to Import!</h4></div>
                     <div ng-hide="success || errors">Click the button below to begin the import of {{ layer.feature_count | number }} features from your {{ layer.file_type }}.</div>
                     <div ng-if="layer.feature_count > 70000 && processing">It looks like your data contains quite a few features. Awesome! This import may take awhile. We'll email you when it is ready for use. Feel free to stick around but it is safe to close this modal and continue using the site.</div>
-                    <div class="step-title" ng-show="success">Congratulations! Click below to view your new Layer.</div>
+                    <div class="step-title" ng-show="success">Congratulations! Click below to publish your new Layer.
+
+                    If you opt not to publish your layer, it will not appear when searching. You can publish the layer by going to your profile page, finding the layer in the Layer tab of your resources, and navigating to the Layer's detail page. There you can edit the metadata and set "Is Published" to be True.</div>
                     <div ng-show="errors">We're sorry but our import service is currently unavailable.  Please try again later.</div>
                     <div class="row" ng-show="success">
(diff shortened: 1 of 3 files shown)


exchange/static/osgeo_importer/partials/uploadWizard.html
```diff
@@ -163,11 +163,13 @@ <h3 align="center"><strong>Layer Metadata</strong></h3>
                     <div class="step-title" ng-hide="success"><h4>Ready to Import!</h4></div>
                     <div ng-hide="success || errors">Click the button below to begin the import of {{ layer.feature_count | number }} features from your {{ layer.file_type }}.</div>
                     <div ng-if="layer.feature_count > 70000 && processing">It looks like your data contains quite a few features. Awesome! This import may take awhile. We'll email you when it is ready for use. Feel free to stick around but it is safe to close this modal and continue using the site.</div>
-                    <div class="step-title" ng-show="success">Congratulations! Click below to view your new Layer.</div>
+                    <div class="step-title" ng-show="success">Congratulations! Click below to publish your new Layer.
```
travislbrundage: Does this not coincide with the upload wizard? This is the only place I've found which matches what is seen in Exchange, but my changes to it aren't taking, and I don't understand why.

travislbrundage:
Getting some really weird behaviour now... this initially works but over time seems like the data is liable to disappear for some unknown reason.

arhote:
Is it becoming unpublished again somehow? or is it just gone gone?

travislbrundage:
Not unpbulished. Just elasticsearch seems to "forget" about it. I am attempting to resolve why this happens or if it's completely unrelated.

arhote:
Potentially related to the "Disappearing Content" issue we were seeing on GVS a while back? Maybe a repro finally?

travislbrundage:
Seems to be unrelated - the common denominator for this issue seems to be logging in and out. I haven't connected anything more than that.

exchange/static/osgeo_importer/partials/uploadWizard.html
travislbrundage: You have to hard clear the cache or incognito. That's why. :upside_down_face:

travislbrundage:
The second elasticsearch issue that is happening now is unrelated to my changes - it happens on master branches/exchange/1.3.0. What seems to happen is that upon logout, one of the layers you most recently saved (thus indexing in elasticsearch) will become unindexed for magical reasons.

But wait, there's more! Additionally, when I log in to a different user, yet another layer will be unindexed via supernatural forces.

In other words, unrelated to my change. My change does not fix that issue, but it doesn't hurt it either.

exchange/static/osgeo_importer/partials/uploadWizard.html
arhote: Yup, that'd do it. (Chrome has a cool disable cache button in the networking tab of dev tools, I like to keep that open when I test for exactly that reason)

PR #380, opened 2017-09-25
Updating profile detail page to work correctly with unified search an…

travislbrundage:
…d provides a way to find your own unpublished layers

### DIFF
--- a/exchange/templates/people/_profile_filters.html
+++ b/exchange/templates/people/_profile_filters.html
@@ -1,0 +1,12 @@
+{% load i18n %}
+
+<ul class="nav nav-pills filter pull-left" id="sort">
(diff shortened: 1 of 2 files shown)


exchange/templates/people/_profile_filters.html
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
```
travislbrundage: Should these changes be done in the GeoNode repo instead? https://github.com/boundlessgeo/geonode/blob/exchange/1.3.0/geonode/people/templates/people/_profile_filters.html

bitner: i think this is fine for exchange rather than geonode

PR #381, opened 2017-09-25
Add unit test for layer_publish

travislbrundage:
Unit test for the layer_publish view which was recently added: https://github.com/boundlessgeo/exchange/pull/374/files#diff-0f1f903cd2c95beaee07a5f4720a4821R137

### DIFF
--- a/exchange/tests/views_test.py
+++ b/exchange/tests/views_test.py
@@ -123,6 +123,30 @@ class LayerMetadataDetailTest(ViewTestCase):
     def test(self):
         self.doit()
 
+
+class LayerPublishTest(ViewTestCase):
+
+    def setUp(self):
+        super(LayerPublishTest, self).setUp()
+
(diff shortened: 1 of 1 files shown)


exchange/tests/views_test.py
```diff
@@ -123,6 +123,24 @@ def setUp(self):
...
+
+    def setUp(self):
+        super(LayerPublishTest, self).setUp()
+
+        from geonode.layers.utils import file_upload
+        self.layer = file_upload(
+            os.path.join(TESTDIR, 'test_point.shp'),
+            name='testlayer'
+        )
+        self.url = '/layers/geonode:testlayer/publish'
```
travislbrundage: Unsure if this should be `geonode:testlayer` or just `testlayer`, but the other tests indicate it should be `geonode:testlayer` and the `/publish` endpoint follows the same conventions.

PR #383, opened 2017-09-29
Fixed broken featured content

davisc:
This PR fixed the bug where featured content failed to load.


travislbrundage:
One quick question about the `||` in the javascript, but otherwise :+1:

davisc:
The response dictionary appears to have a different payload based on how Exchange is configured so to accommodate both, that syntax allows you to use the logical OR to return which ever is defined or return undefined if neither.

PR #400, opened 2017-10-19
Added instruction for geonode branch

elpaso:

PR #404, opened 2017-10-23
updating readme & dev setup

mrcnc:
Updates the root readme to point to the readme in the `docker` folder for local environment setup.   I also made some documentation updates in the docker readme that supersede #400 and #401

Additionally, I changed the port and removed the configuration exposing port 443 from the exchange container.  This fixes an issue that prevented the proxy container from starting.  Now you can access the Exchange UI through the proxy at http://localhost and https://localhost

### DIFF
--- a/docker/README.rst
+++ b/docker/README.rst
@@ -109,8 +109,6 @@ You can use pip to install the Python dependencies for the Exchange Django app::
 
     pip install -r requirements.txt
 
-If you are using Postgres 10, you will need to update the psycopg2 dependency to 2.7.3 or higher
-
 
 Preparations
 ------------
--- a/docker/README.rst
+++ b/docker/README.rst
@@ -22,46 +22,6 @@ Docker
     e.g. by putting your user in the docker group. Other times, you will have to perform such steps yourself, according to the instructions,
     then log out and back in. In some environments, such as Arch Linux, you may need to manually start the docker service after installing Docker.
 
-Python 2.7
-~~~~~~~~~~
-
-You also need to setup a Python 2.7 development environment (Python 3 is not supported yet).
-
-You can do this by setting up a new virtual environment::
-
-  cd /path/to/exchange
-  pip install virtualenv
-  virtualenv .venv
-  source .venv/bin/activate
-
-GDAL 2
-~~~~~~
-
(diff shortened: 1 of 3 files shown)


docker/README.rst
```diff
@@ -7,121 +7,135 @@ features to reduce pains we've had working with our Vagrant config, such as
...
-   where on your computer to find the directories which will be mounted inside
-   the containers as `/mnt/exchange`, `/mnt/geonode` and `/mnt/maploom`. If
-   these are not set accurately, then the exchange container and the Celery
-   worker container will not be able to start properly because they won't have
-   code to start with.
+    pip install -r requirements.txt
 
-   Using host directories like this isn't a Docker thing, just a convenience
-   carried over from the Vagrant dev config.
+If you are using Postgres 10, you will need to update the psycopg2 dependency to 2.7.3 or higher
```
mrcnc: When #402 is merged, we can remove this line

docker/README.rst
```diff
@@ -7,131 +7,133 @@ features to reduce pains we've had working with our Vagrant config, such as
...
+    e.g. by putting your user in the docker group. Other times, you will have to perform such steps yourself, according to the instructions,
+    then log out and back in. In some environments, such as Arch Linux, you may need to manually start the docker service after installing Docker.
 
-   Depending on platform, Docker's install instructions and packages sometimes
-   set things up so that you do not have to use `sudo docker`, e.g. by putting
-   your user in the `docker` group. Other times, you will have to perform such
-   steps yourself, according to the instructions, then log out and back in.
-   In some environments, such as Arch Linux, you may need to manually start the
-   docker service after installing Docker.
+Python 2.7
```
arhote: When using the docker environment, it's not necessary to setup a virtual environment, install gdal locally, or pip install exchange's requirements.txt, as those are all done inside the container as necessary. As such, I'd suggest either moving those sections out of the docker readme, or move them to an appropriate section, like "If you want to configure exchange locally."

mrcnc: Ok...that makes sense.  I'll remove it for now and make another PR to add the "configuring Exchange locally" after I have some more experience developing & debugging Exchange

arhote: Sounds perfect, Thanks!

PR #405, opened 2017-10-23
update images to use the name:tag in the container registry

mrcnc:
According to [the docs](https://docs.docker.com/compose/compose-file/compose-file-v2/#build) if you are specifying the `build` and `image`, then the image will be built using the image-name:tag specified for the `image` key for that service.

This makes it easier to push the latest version of a container to the container registry by using the service name.  For example,`docker-compose push geoserver` will push the latest built version of the geoserver container to `quay.io/boundlessgeo/exchange-geoserver`

Additionally, if you don't want to build the latest containers for each of the dependent services (b/c it takes a long time), you can comment out the `build` line for each service, then docker will pull the latest image from the registry.


davisc:
We should consider pushing containers to the registry as part of the ci pipeline. @camerontesterman @amirahav thoughts?

walker:
@davisc @camerontesterman @amirahav Is there a way in quay to differentiate between release containers & dev (bleeding edge) containers?

amirahav:
The idea is that there would be two “name spaces” - bex and bex-dev.
bex-dev:latest will always be the latest image pushed by CI on a successful build, while bex:latest will be the latest stable release. bex will also have version specific images, so bex:1.3.0, bex:1.3.1, bex:1.4.0, etc

mrcnc:
I'd prefer to use tags to differentiate between the different releases or development versions.  You can tag an image based on the GitHub branch name `boundlessgeo/bex:master` or you could create a convention that `boundlessgeo/bex:latest` is the image built from master branch.  I don't see an advantage to using another image when the same thing could be done using tags. 

I'm closing this PR b/c it has been superseded by #413

amirahav:
I don't like the idea of having ```bex:dev```, ```or bex:master```, instead I think that ```bex``` should be reserved to released/taged versions, where bex:latest is the latest tagged version, and then ```bex-dev:whatever``` can be used with branch names, etc.

But I'm open to this being too confusing....

mrcnc:
Ok...it's just a personal preference that I have.  I look at an image in a container repository the same way that I look at a repository on GitHub.  I wouldn't create a new repo for the released versions of a project on GitHub, so I wouldn't do that for a container repo.  If you wanted to have the convention be `bex:latest` is the latest release, that makes sense to me.

amirahav:
Just a difference in how we look at an image in a repository :-)
When creating branches we need to make sure we don't name the branch with the full version, so instead of ```1.4.1``` it should be named ```1.4.x```, (which we should do anyway....).
Then ```bex:latest``` should be latest tagged release, and we can either do ```bex:1.4.1``` or ```bex:release-1.4.1``` for specific released versions. (I prefer the first one)

PR #413, opened 2017-10-27
Use geoserver 2.12. docker image names

amirahav:
Uses GS 2.12. note - gs-geofence and geofence-server are somewhat problematic with the restlet to spring mvc refactoring done upstream. There are a couple of variables being set even though they shouldn't be.

Should replace https://github.com/boundlessgeo/exchange/pull/405 for container naming convention.
There is no need to push containers to quay before merging this pr.

Should be tested before merging.


camerontesterman:
Tested this configuration. Unable to upload layers, tells me that the upload failed.

amirahav:
@camerontesterman thank you for testing this!
It looks like there are a few other things that upstream broke, and ```/geoserver/rest``` is returning a 406.

amirahav:
This should now work

camerontesterman:
great thank you - in process of reviewing now

PR #414, opened 2017-10-30
Fix thumbnail generation and missing post_save steps in celery workers.

jdgarrett:
- Move thumbnail creation to an asynchronous celery worker
- Add .env to worker container to ensure that it properly updates the index on save
- Update thumbnail endpoint to return a missing thumbnail image instead of a blank gif
- Add a step to the import process to trigger a final save so that thumbnail gets generated after the import process completes


bitner:
This PR looks good. I'm going to go ahead and merge this, but we should definitely get some tests added to cover these tasks.

exchange/thumbnails/tasks.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+    tries = 0
+    while tries < 30:
+        resp, image = http_client.request(thumbnail_create_url)
+        if 200 <= resp.status <= 299:
+            if 'ServiceException' not in image:
+                return image
+        else:
+            # Unexpected Error Code, Stop Trying
```
arhote: It might be a good idea to at least log the response status and information for debugging purposes in the future. Even if we don't care if a user sees the information, it'd be helpful to devs.

PR #420, opened 2017-11-03
Registry Docker fix - added Registry documentation

camerontesterman:
added a little blurb about getting registry working in the docker environment. Fixed some environment variables - managed remote services work now and don't throw a celery error.


docker/README.rst
```diff
@@ -190,6 +190,20 @@ GeoServer can be browsed at
...
+
+You can access Registry at
+
+    http://172.16.238.2/registry
+
+Prior to using Registry, you will need to create the Registry catalog. You must
+do this after the container is running in order to use Registry.
+To do this, run the following command in your terminal.
+
+    curl -XPUT http://172.16.238.2/registry/catalog/registry/csw
```
arhote: Is there not a way we can put this in the registry Dockerfile or something so that users aren't having to manually run a curl every time they bounce their container? That seems like something that will get overlooked real quick.

arhote: Follow up: After discussion w/Cameron we agreed to Parking Lot this at standup to get the whole team's opinion on the best path forward.

PR #422, opened 2017-11-03
compatibility with elasticsearch-dsl 5.3.x

amirahav:
compatibility with elasticsearch-dsl 5.3.x aggregations is ```AggResponse``` and not a dict

### DIFF
--- a/exchange/views.py
+++ b/exchange/views.py
@@ -400,7 +400,7 @@ def unified_elastic_search(request, resourcetype='base'):
     # with overall count and any display name or icon that should be used in UI
     aggregations = overall_results.aggregations
     facet_results = {}
-    for k in aggregations:
+    for k in facet_fields:
         buckets = aggregations[k]['buckets']
         if len(buckets)>0:
             lookup = None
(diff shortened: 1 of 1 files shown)


exchange/views.py
```diff
@@ -400,7 +400,7 @@ def field_name(field, mappings):
     # with overall count and any display name or icon that should be used in UI
     aggregations = overall_results.aggregations
     facet_results = {}
-    for k in aggregations:
+    for k in facet_fields:
         buckets = aggregations[k]['buckets']
```
travislbrundage: This line should be `buckets = k.buckets`

exchange/views.py
```diff
@@ -555,7 +555,7 @@ def field_name(field, mappings):
 
     # get facets based on search criteria, add to overall facets
     aggregations = results.aggregations
-    for k in aggregations:
+    for k in facet_fields:
         buckets = aggregations[k]['buckets']
```
travislbrundage: This line should be `buckets = k.buckets`

exchange/views.py
amirahav: The reason that wasn’t done is in multiple other places in the function the entire dict of that bucket would then be used as the key name.
If you’re worry here is a mismatch between facet_fields and the keys in elastic, we can call dict() on aggregations.
Would you prefer that?

PR #433, opened 2017-11-14
Fixed BEX-417 by removing the extraneous non-displayed categories

walker:
Fixed BEX-417 by removing the extraneous non-displayed categories that were just matched...

This will **_not_** fix existing installations. Existing installations will need to be edited via the django admin. ({host}/admin/base/topiccategory/)


walker:
@arhote since @bitner isn't here...maybe check that second one...but I see the problem in the faceting code...

PR #440, opened 2017-11-17
REGISTRY_SERVER_URL

walker:
"...to get it to work you have to set the `REGISTRY_SERVER_URL` incorrectly"


walker:
@camerontesterman this doesn't seem to have completely fixed registry (remote service publish action) on a clean local instance.

PR #445, opened 2017-11-28
Enable LOOM_STYLING_ENABLED

davisc:
This PR enables the LOOM_STYLING_ENABLED setting by default for MapLoom styling.


theduckylittle:
@davisc  - Does this add styling when you rebuild the exchange container? In order to get styling enabled I needed to actually has the `LOOM ... = True` to my default.py settings file.

davisc:
@theduckylittle see following in the default.py

# MapLoom Styling Control
LOOM_STYLING_ENABLED = str2bool(os.getenv('LOOM_STYLING_ENABLED', 'False'))

theduckylittle:
:+1: Thanks for the clarification.

PR #448, opened 2017-11-30
Bex 330 331

travislbrundage:
Addresses BEX-330 and BEX-331. Requires: https://github.com/boundlessgeo/geonode/pull/94

That simply bumps the elasticsearch version down in GeoNode as well, which seems to override my change in Exchange. The version changing does not seem to have any affect, but it is more proper to have our python library versions of elastic match the actual elasticsearch backend version we are using, which is 2.x.


exchange/elasticsearchapp/management/commands/clear_index.py
```diff
@@ -0,0 +1,21 @@
...
+    help = "Clears out the search index completely."
+
+    def handle(self, **options):
+        '''
+        Clears out the search index completely.
+        '''
+        self.stdout.write("Removing all documents from your index.")
+        es = Elasticsearch(settings.ES_URL)
+        # Iterate through every index in our elasticsearch endpoint
+        for index in es.indices.get_mapping().iteritems():
```
bitner: We definitely shouldn't be just clearing *all* indices that may be on a an ES instance.

exchange/views.py
```diff
@@ -354,42 +348,29 @@ def newDSLBaseInit(self, _expand__to_dot=False, **params):
         filter_set_ids = map(str, filter_set.values_list('id', flat=True))
         # Do the query using the filterset and the query term. Facet the
         # results
-        q = Q({"match": {"_type": "layer"}})
```
bitner: This is in here to be able to get all registry layers which do not have permissions

exchange/views.py
```diff
@@ -336,11 +331,10 @@ def newDSLBaseInit(self, _expand__to_dot=False, **params):
     # This provides the overall counts and all fields for faceting
 
     # only show registry, documents, layers, stories, and maps
-    q = Q({"match": {"_type": "layer"}}) | Q(
```
bitner: This is needed to get registry layers which come in with _type=layer

bitner:
Hrmm, looks like this brings our views.py coverage down quite a bit as well.

travislbrundage:
The code coverage decrease might be because I apparently removed a test accidentally somehow. Adding that back in. Will also add back in the type check for registry layers and rebase on master.

exchange/elasticsearchapp/management/commands/clear_index.py
travislbrundage: What should a clear_index command do otherwise? To note, I was just emulating the django-haystack command. Thus, the new commands are run the same way as the previous ones and have the same affect. Happy to change it if there's a better way to do things.

exchange/layers/models.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def prepare_subtype(self):
+        if self.storeType == "dataStore":
+            return "vector"
+        elif self.storeType == "coverageStore":
+            return "raster"
+        elif self.storeType == "remoteStore":
+            return "remote"
+        else:
+            return None
```
travislbrundage: Unsure if returning `None` here is appropriate or if it should be an empty string - however, I don't think the `else` condition should ever be possible to occur.

exchange/templates/search/search.html
```diff
@@ -22,7 +22,7 @@ <h2>{% trans "Search" %} <span ng-if="text_query != ''">:</span> <span ng-bind="
     {% include "_permissions_form_js.html" %}
   {% endif %}
   <script type="text/javascript">
-    SEARCH_URL = '{% url 'api_get_search' api_name='api' resource_name='base' %}'
+    SEARCH_URL = '{% url 'unified_elastic_search' resourcetype='base' %}'
```
travislbrundage: This was changed to match the URL's specified in our `urls.py`: https://github.com/boundlessgeo/exchange/blob/master/exchange/urls.py#L85

I'm not sure exactly what the previous link did, but I believe it linked to here: https://github.com/boundlessgeo/geonode/blob/exchange/1.2.0/geonode/api/resourcebase_api.py#L535

This no longer works with the elasticsearch refactor, though I am not 100% sure why. I believe the code it was linking to was doing something with django-haystack, then redirecting to the unified_elastic_search view. Because django-haystack has been removed, I think the reference is no longer relevant and we can just link to the unified_elastic_search view.

exchange/documents/__init__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
travislbrundage: These are the standard headers, should they be like this? Should I add one to the elasticsearchapp as well?

exchange/layers/models.py
arhote: Couldn't self.storeType possibly be `None` which would trigger the else?

travislbrundage:
There are three issues remaining with this pull request, one major and two minor.

Minor: I noticed that Documents were not indexing categories previously, even though that is a field they have. Should I make the Documents index categories in elasticsearch?

Also, our people and group search pages don't appear to be using elasticsearch right now, but the django api directly. Should I update these endpoints as well to use elasticsearch? It will require a similar function as the unified_elastic_search view but customized for people/groups respectively instead of just resources.

Major: Loading the models in Exchange so we don't have to make modifications to our geonode **currently does not work**. The reason for this is that it causes a model conflict with registry. The stack trace is given below. I am unsure how to fix this, but I've done testing by putting the changes in geonode and things seem to work that way. Related, there are quite a few places where the geonode versions of the models are referenced - I presume these should all be modified to reference exchange instead?

Stacktrace:
```
exchange     | Traceback (most recent call last):
exchange     |   File "/env/lib/python2.7/site-packages/django/utils/autoreload.py", line 229, in wrapper
exchange     |     fn(*args, **kwargs)
exchange     |   File "/env/lib/python2.7/site-packages/django/core/management/commands/runserver.py", line 107, in inner_run
exchange     |     autoreload.raise_last_exception()
exchange     |   File "/env/lib/python2.7/site-packages/django/utils/autoreload.py", line 252, in raise_last_exception
exchange     |     six.reraise(*_exception)
exchange     |   File "/env/lib/python2.7/site-packages/django/utils/autoreload.py", line 229, in wrapper
exchange     |     fn(*args, **kwargs)
exchange     |   File "/env/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
exchange     |     apps.populate(settings.INSTALLED_APPS)
exchange     |   File "/env/lib/python2.7/site-packages/django/apps/registry.py", line 115, in populate
exchange     |     app_config.ready()
exchange     |   File "/mnt/exchange/exchange/elasticsearchapp/apps.py", line 8, in ready
exchange     |     import exchange.elasticsearchapp.signals
exchange     |   File "/mnt/exchange/exchange/elasticsearchapp/signals.py", line 1, in <module>
exchange     |     from exchange.layers.models import Layer
exchange     |   File "/mnt/exchange/exchange/layers/models.py", line 66, in <module>
exchange     |     class Style(models.Model):
exchange     |   File "/env/lib/python2.7/site-packages/django/db/models/base.py", line 309, in __new__
exchange     |     new_class._meta.apps.register_model(new_class._meta.app_label, new_class)
exchange     |   File "/env/lib/python2.7/site-packages/django/apps/registry.py", line 221, in register_model
exchange     |     (model_name, app_label, app_models[model_name], model))
exchange     | RuntimeError: Conflicting 'style' models in application 'layers': <class 'geonode.layers.models.Style'> and <class 'exchange.layers.models.Style'>.
```
Note that if you remove any of the models, it just conflicts with another and will also conflict on the base models themselves. Unsure why it conflicts on Style model first.

exchange/elasticsearchapp/management/commands/clear_index.py
bitner: it should clear out the indices that it created, if this was rolled out on an instance of elastic that was used by something else, this could do "bad things"

exchange/templates/search/search.html
bitner: unified search should be using the same url endpoints and just overriding them via urls.py, but as we are full in for always using unified search, this is a good change

exchange/elasticsearchapp/management/commands/clear_index.py
travislbrundage: Okay, that's fine. So should I just clear the `layer-index`, `map-index`, etc. that is added by elasticsearchapp specifically? That way it will only be deleting indexes that our app created. The only downside of this is that I don't know how to make it programmatic that way. I would just literally reference them by their name, i.e. `layer-index`, for every one of them.

travislbrundage:
Major issue should be resolved now by putting the model changes in https://github.com/boundlessgeo/geonode/pull/94

The code coverage drop is occurring because the unified search tests are failing due to import not working correctly. If I reference data that is already on my Exchange instance rather than trying to do the import in the test, all the tests will pass, and the code coverage should remain roughly stable. Thus, the issue should be mostly localized to its `setUp`, fixing it should bring the coverage back up (to note, the same issue appears with the osgeo importer test and should have the same fix).

There is also a small drop for the elasticsearchapp management commands, but I am not sure how those ought to be tested. The actual backend has a test which should cover what the app primarily does.

exchange/tests/views_test.py
```diff
@@ -217,13 +222,47 @@ def test(self):
...
+    def test(self):
+        self.doit()
+
+
 @pytest.mark.skipif(settings.ES_UNIFIED_SEARCH is False,
                     reason="Only run if using unified search")
 class UnifiedSearchTest(ViewTestCase, UploaderMixin):
 
+    # TODO: Creation of all object is very finnicky for some reason
     def setUp(self):
```
travislbrundage: This, and/or the `upload_files` function is what is causing all the tests here to fail. What I need help with is making the mock data to get ingested in elasticsearch. If I do this part manually, thus skipping the `setUp` and referencing data I put in myself, all the tests will pass. Instead of having to take my word for it though we should get this fixed. I could not find any examples of how to do this correctly, so any feedback here is appreciated greatly.

exchange/layers/models.py
travislbrundage: If it was possible to get into that state, yeah. But that shouldn't be possible. `storeType` is how the layer is being stored, so if it's None.. the layer isn't stored... which doesn't make any sense and shouldn't be possible if the Layer model was created.

travislbrundage:
I've added a test for the `clear_index` and `rebuild_index` management commands.

Remaining things to note: I don't know how I can test `haystack_clear` as a command when its purpose is to clear our the haystack index due to an upgrade.

Related to that, I want to make sure an upgrade is tried out for these changes (geonode changes are here: https://github.com/boundlessgeo/geonode/pull/94 
chef script updates are in the exchange rpm cookbooks). What we want/expect is for a smooth transition from the old haystack method of using elasticsearch to the new method of using elasticsearch. All data and functionality should be preserved such that the user does not notice anything. However, on the backend, haystack's `modelresult` index should be cleared from their elasticsearch, and there should be new indices corresponding to our models which are able to be searched/faceted on just like before.

davisc:
@travislbrundage with this change, lets replace `ES_UNIFIED_SEARCH` with `ES_SEARCH`

travislbrundage:
Done ^

davisc:
Since Geonode also depends on the elasticsearchapp defined in Exchange core, have we thought about pulling that app out into its own project?

bitner:
@davisc I think that makes a lot of sense, but possibly a priority for after this release

travislbrundage:
New PR here: https://github.com/boundlessgeo/exchange/pull/468

PR #452, opened 2017-12-06
Expose option to set permissions on the layer.

davisc:
screen shot 2017-12-06 at 7 06 57 am

depends on https://github.com/boundlessgeo/geonode/pull/96

### DIFF
--- a/exchange/static/osgeo_importer/importer.js
+++ b/exchange/static/osgeo_importer/importer.js
@@ -2,605 +2,602 @@
 
 (function() {
 
-  var httpService_ = null;
-  var layerService_ = null;
-  var q_ = null;
+    var httpService_ = null;
+    var layerService_ = null;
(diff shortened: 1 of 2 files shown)


exchange/static/osgeo_importer/importer.js
```diff
@@ -2,605 +2,602 @@
...
-    };
+                } else {
+                    if (layer.configuration_options.version_control === true) {
+                        layer.configuration_options.geoserver_store = {'type': 'geogig'};
+                    } else {
+                        delete layer.configuration_options.geoserver_store;
+                    }
+                }
+                if (layer.configuration_options.public == false) {
+                    layer.configuration_options.permissions = {};
```
davisc: @amirahav made a client-side change to clear the perms dict if public is set to false via the wizard.

PR #453, opened 2017-12-06
add capabilities endpoint

mrcnc:
The purpose of this PR is to add a new endpoint for clients to discover the capabilities of an Exchange instance.  It's like an about page but meant for code instead of humans.  

I've added a new endpoint `/capabilities` that returns a JSON response.  Also, I made it accessible without authentication so that the Anywhere app can determine if Exchange is configured with the mobile extension.  Thanks to @harts-boundless for implementing the initial idea.

Here's an example of the response:
```
{
    "capabilities": {
        "mobile": false,
        "site_name": "example.com",
        "versions": {
            "exchange": {
                "commit": "7c99b46",
                "version": "1.4.dev"
            },
            "geonode": {
                "commit": "79aa25b",
                "version": "2.6.1.9"
            },
            "geoserver": {
                "commit": "f15bed7",
                "version": "2.12-SNAPSHOT"
            }
        }
    }
}
```


walker:
@mrcnc, can `"maploom": true,` instead be `"default_map_viewer":"maploom"` or some such thing?

walker:
Also, what's `"slack"`?

walker:
@mrcnc Also, you'll note the drop in code coverage caused by this PR. You'll likely want to reverse that drop before this'll get merged.

davisc:
@walker slack is Slack... send notifications from Geonode to a slack channel.

walker:
@davisc lol. I had no idea.

theduckylittle:
I think spec'ing a default map client, as per @walker 's suggestion is a better starting point as the _current_ version of exchange can have a litany of map clients and enumerating all of them probably doesn't provide all that much useful information.  In the future, hopefully, we will have 'the one mapping client to rule them all'.

I'd also like to point to this: https://github.com/boundlessgeo/exchange/blob/sdk-dev/exchange/castling/views.py#L15

Where we have the ability to get the current auth token for when making AJAX calls.  It also points to the relative GeoServer path.

mrcnc:
@theduckylittle @walker I've updated the capabilities to include the requested `default_map_viewer` property.  I'm not sure how to determine the other map viewers besides Maploom so currently the property will be an empty string if the maploom app isn't installed.

walker:
> I'm not sure how to determine the other map viewers besides Maploom so currently the property will be an empty string if the maploom app isn't installed.

We can improve this later, I think.

mrcnc:
I rebased and updated the PR to merge into the 1.3.x branch

amirahav:
Let's merge into master and backport instead....

mrcnc:
Ok....changed it back to master

davisc:
@mrcnc is all of the information included in the response needed for mobile? Would it make sense to reduce it down to the specific bits?

amirahav:
@mrcnc what if we limited only to the following:

```
{
	"capabilities": {
		"mobile": false,
		"versions": {
			"exchange": {
				"commit": "c764c1c",
				"version": "1.3.1dev"
			},
			"geonode": {
				"commit": "a6fc931",
				"version": "2.6.1.a"
			},
			"geoserver": {
				"commit": "f15bed7",
				"version": "2.12-SNAPSHOT"
			}
		}
	}
}
```

Does mobile need anything else?

mrcnc:
Updated the response based on feedback from @davisc and @amirahav

amirahav:
@mrcnc let me know when these has been removed:
https://github.com/mrcnc/exchange/blob/bd6a2cdfce5ee15ce656cea5c1c8b340adfee758/exchange/views.py#L18-L35

And these two have been updated:
https://github.com/mrcnc/exchange/blob/bd6a2cdfce5ee15ce656cea5c1c8b340adfee758/exchange/views.py#L41
https://github.com/mrcnc/exchange/blob/bd6a2cdfce5ee15ce656cea5c1c8b340adfee758/exchange/views.py#L50

And we can merge the PR

davisc:
@mrcnc If you add the site name to the response, you could used it in the app as an alternative display for the url in the header bar.

mrcnc:
@davisc I added the `site_name` property.
@amirahav I refactored the `views.py` to what I think you were looking for.

PR #467, opened 2017-12-12
the 'no image' gif is larger than blank gif used before

bitner:
It appears as though the thumbnail has been changed from blank to a 'no image' gif. This updates the test to look for the appropriate size.


walker:
@bitner Is there no way to get an auto-render of the map as a thumbnail _without_ going to "Edit Layer" -> "Set"?

PR #476, opened 2017-12-15
BEX-502 fixed card display

davisc:
Fixed:

- Link not associated with the title of the card
- The layer data type should say FeatureService
- The MapServer link is not showing up

### DIFF
--- a/exchange/static/js/search/search.js
+++ b/exchange/static/js/search/search.js
@@ -20,7 +20,7 @@
         return function (text) {
 
             var mapping = {'AIMS--http-get-feature': 'FeatureServer', 'AIMS--http-get-map': 'MapServer',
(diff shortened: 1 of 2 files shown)


exchange/static/js/search/search.js
```diff
@@ -20,7 +20,7 @@
         return function (text) {
 
             var mapping = {'AIMS--http-get-feature': 'FeatureServer', 'AIMS--http-get-map': 'MapServer',
-                'AIMS--http-get-image': 'ImageServer'}
+                'AIMS--http-get-image': 'ImageServer', 'ESRI:ArcGIS:FeatureServer': 'FeatureServer'}
```
travislbrundage: Does this work with the split below on `friendlyParts` of the `text`? https://github.com/boundlessgeo/exchange/blob/9d866b9eefcf19ab747171b3c2a0f4bc1d27c1a3/exchange/static/js/search/search.js#L29

PR #487, opened 2017-12-18
Uploader test fixes

bitner:
Fixes issues with Exchange Tests, particularly issues that were caused with tests that did any uploading. This moves all uploading logic into the UploaderMixin in osgeo_importer_upload_test.py and cleans up. There are three geogig tests that are skipped right now due to needing to figure out some issues with geogig setup and one test that is skipped because we need to add temporal extent setup to the base upload for search and two tests that are skipped that relate to the bug that has been reported by Ami that raster uploads receive the name referencing the entire file path.


bitner:
rejiggered in pr 489

PR #495, opened 2017-12-22
Include exchange docs in exchange

BerryDaniel:
This addition adds the exchange documentation to the exchange repo. It does add a submodule boundlessgeo/sphinx-theme. An addition was added to setup.py to now have a command `python setup.py build_sphinx`. The intent is to include the necessary packages in requirements.txt and run the command above, before the collectstatic step. This will now have the documentation verified that it builds correctly in the exchange pipeline during the `Set-Up` stage.

- sphinx was forked and was modified to include an option to read an os environment for a specific build path, in this case, the build path is the exchange/static/docs


davisc:
Looks good.

PR #497, opened 2018-01-02
Overwrite django admin default header

davisc:
As of Django 1.7 you don't need to override templates. You can now implement site_header, site_title, and index_title attributes to easily change the admin site’s page title and header text.

screen shot 2018-01-02 at 9 41 57 am

screen shot 2018-01-02 at 9 45 33 am


bitner:
@BerryDaniel it looks like the jenkins error is with a container not being up yet, can you check the build?

davisc:
I'm going to move this change over to Geonode.

PR #505, opened 2018-01-05
Fix setUp and tearDown for elasticsearch tests - call clear and rebui…

travislbrundage:
…ld index commands to work with a clean index


amirahav:
https://github.com/boundlessgeo/exchange/pull/509 solves this issue

PR #507, opened 2018-01-05
Add regions to facet search

travislbrundage:
In the future, it may be useful to add `regions` to the fields we can search on as in the elasticsearch 6.x branch, here: https://github.com/boundlessgeo/exchange/blob/elasticsearch-6.x-compat/exchange/search/views.py#L224

However GVS requested to be able to filter by the region on the search page, hence this change.


bitner:
@walker should chime in here as to what we want by default in exchange, this can be overridden for GVS using the ADDITIONAL_FACETS setting

travislbrundage:
Gotcha - will update their settings for this case then. Thanks!

PR #508, opened 2018-01-05
Bump geonode and maploom submodules

amirahav:


amirahav:
This failure is caused by https://github.com/boundlessgeo/geonode/pull/109.
Closing until it's fixed.

PR #512, opened 2018-01-05
Adding exchange qa members to auto review as code owners in github

BerryDaniel:
https://github.com/blog/2392-introducing-code-owners


boundlessssengupta:
@BerryDaniel Will it require all members of the bex-qa team to review or anybody from that group?

BerryDaniel:
should just be one member, we can test it out :)

PR #516, opened 2018-01-08
fixture to add bcs-hosted-data Remote Service

iant1212:
Adds a fixture to pre-populate Boundless hosted data.  entrypoint.sh runs loaddata if $DEV==True


mrcnc:
I see that you are using primary keys in the fixtures but I'm not sure that's what we want b/c it can potentially override data that already exists.  While I understand that this fixture is only being loaded for development, this data might be something we want to add to non-development instances of Exchange used for trials or demo purposes.  I'll defer to @BerryDaniel or others more knowledgable on best practices for loading datasets into Exchange

BerryDaniel:
You could use `pk: null`, but everytime you restart exchange or do a `docker-compose up` wouldn't it create a duplicate entry? It might be best to create a `pk: 9999` instead.

davisc:
Or we could use the import service management command. `python manage.py importservice`

iant1212:
The 'importservice' command seems to be the way to go as we can bypasses hardcoded PKs, and not have to deal with fixture files at all.  Although to get it to test without error I had to change some things in the geonode/geonode/services/management/commands/importservice.py

staceyberry:
@iant1212 the branch has conflicts and must be resolved before we can test it.

docker/exchange/entrypoint.sh
```diff
@@ -7,6 +7,57 @@ maploom_html='/usr/local/lib/python2.7/site-packages/maploom/templates/maps/mapl
...
+  fi
+  if [ "$plugins" ]; then
+    ADDITIONAL_APPS=$(IFS=,; echo "${plugins[*]}")
+  fi
+  $manage runserver 0.0.0.0:8000
+else
+  pip freeze
+  C_FORCE_ROOT=1 celery worker --app=exchange.celery_app:app -B --loglevel DEBUG
+fi
+=======
```
amirahav: @iant1212 can you remove this commit message?

PR #517, opened 2018-01-10
Combine worker and exchange container

BerryDaniel:
- run celery and django using supervisord
- set log output for supervisord to stdout
- add startup env check for DEV=True, if set then django runserver is used, else waitress with DEBUG=False
- updated README


BerryDaniel:
@bitner passing now

PR #555, opened 2018-01-30
BEX-635

BerryDaniel:
## JIRA Ticket
BEX-635

## Description
Validate if a migration needs to run. Added the check for Jenkinsfile and Makefile `make test`.
- added migrations for account and user_messages and configured them as MIGRATION_MODULES - https://docs.djangoproject.com/en/1.8/ref/settings/#migration-modules
Note: 3pm = Third Party Modules

screen shot 2018-01-30 at 9 58 16 am

Also modified the following:
- removed the sed check from Dockerfile for exchange replacing geonode. Geonode is installed from the vendor directory, so the entry was just added to requirements.txt
- adjusted pipeline to include parallel stages for style checks (pycodestyle, yamllint and flake8) and build (bex - docker-compose and maploom). 
- Also added the exchange-mobile-extension to be auto added to Jenkins CI - Exchange Pipleine as an installed extension. 
- Adjusted coverage output to terminal only, since we were currently not using the xml file and this will provide immediate feedback in Jenkins logs.

## TODO
- [X] pycodestyle (`make lint`)
- [X] tests (`make test`)

## Steps to Test or Reproduce
_Add detailed steps for testing/review team to verify._

## Setup Environment with PR

1. Cleanup previous state

```bash
make purge
```

2. Checkout PR

__Using git command (command line interface)__
```bash
pr_id=ADD_PR_NUMBER_HERE
git checkout master
git branch -D $pr_id
git fetch
git fetch origin pull/$pr_id/head:$pr_id
git checkout $pr_id
git submodule init
git submodule update --remote --recursive
```

__Using [Gitkraken](https://www.gitkraken.com/)__

+ In GitKraken, right click on the pull request you want to review. Select Add Remote and Checkout (the PR).

3. Start exchange

```bash
make start
```

4. Exchange Healthcheck

```bash
docker inspect --format '{{ .State.Health.Status }}' exchange
```

__NOTE:__ Only continue the following steps if the output from the above command is `healthy`. You may have to wait 
a few minutes.

---
@boundlessgeo/bex-qa


exchange/core/migrations/0010_auto_20180125_1314.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('core', '0009_auto_20170906_0857'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='thumbnailimage',
+            name='thumbnail_image',
+            field=models.ImageField(upload_to=b'/app/data/media/thumbs'),
```
amirahav: @davisc @BerryDaniel We should probably change the models and remove the path. The path can change depending on installation type.

exchange/3pm/account/0001_initial.py
```diff
@@ -0,0 +1,804 @@
...
+        ),
+        migrations.CreateModel(
+            name='EmailAddress',
+            fields=[
+                ('id', models.AutoField(
+                    verbose_name='ID',
+                    serialize=False,
+                    auto_created=True,
+                    primary_key=True)),
+                ('email', models.EmailField(unique=True, max_length=254)),
```
amirahav: We might want to set https://github.com/boundlessgeo/exchange/blob/master/exchange/settings/default.py#L530 to `False` and then change the migration.
In cases where we use multiple authentication types, that attribute will be set to false to allows an account with the same email address to login

PR #558, opened 2018-02-02
Psbex 70

travislbrundage:
## JIRA Ticket
PSBEX-70

## Description
Adds pki route routine and overrides form used by geonode for service registration in order to inject the routine.

## TODO
- [ ] pycodestyle (`make lint`)
- [ ] tests (`make test`)

## Steps to Test or Reproduce
_Add detailed steps for testing/review team to verify._

## Setup Environment with PR

1. Cleanup previous state

```bash
make purge
```

2. Checkout PR

__Using git command (command line interface)__
```bash
pr_id=ADD_PR_NUMBER_HERE
git checkout master
git branch -D $pr_id
git fetch
git fetch origin pull/$pr_id/head:$pr_id
git checkout $pr_id
git submodule init
git submodule update --remote --recursive
```

__Using [Gitkraken](https://www.gitkraken.com/)__

+ In GitKraken, right click on the pull request you want to review. Select Add Remote and Checkout (the PR).

3. Start exchange

```bash
make start
```

4. Exchange Healthcheck

```bash
docker inspect --format '{{ .State.Health.Status }}' exchange
```

__NOTE:__ Only continue the following steps if the output from the above command is `healthy`. You may have to wait 
a few minutes.

---
@boundlessgeo/bex-qa

### DIFF
--- a/exchange/pki/forms.py
+++ b/exchange/pki/forms.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+#########################################################################
+#
+# Copyright (C) 2018 Boundless Spatial
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+#########################################################################
+
+from django import forms
+from geonode.services.forms import CreateServiceForm
+from geonode.services import enumerations
+from geonode.services.serviceprocessors import get_service_handler
+from django.core.exceptions import ValidationError
+from .models import SslConfig
+
+
+def get_ssl_config_choices():
+    SSL_CONFIG_CHOICES = ()
+    for ssl_config_choice in SslConfig.objects.all():
+        SSL_CONFIG_CHOICES += ((ssl_config_choice.pk, ssl_config_choice.name),)
+
+    return SSL_CONFIG_CHOICES
+
+
+class CreatePKIServiceForm(CreateServiceForm):
+    # In addition to normal service registration,
+    # we want to be able to associate it with an SslConfig model
+    ssl_config = forms.ChoiceField(
+        label="SSL Config",
+        choices=get_ssl_config_choices(),
+        initial='',
+    )
+
+    def clean_url(self):
+        # Here we can add our own validation if necessary
+        super(CreatePKIServiceForm, self).clean_url()
+
+    def clean(self):
+        # Unfortunately we have to override and redo the function
+        # In order to inject the pki rerouting
+        url = self.cleaned_data.get("url")
+        service_type = self.cleaned_data.get("type")
+        if url is not None and service_type is not None:
+            try:
+                service_handler = get_service_handler(
+                    base_url=url, service_type=service_type)
+            except Exception:
+                raise ValidationError(
+                    _("Could not connect to the service at %(url)s"),
+                    params={"url": url}
+                )
+            if not service_handler.has_resources():
+                raise ValidationError(
+                    _("Could not find importable resources for the service "
+                      "at %(url)s"),
+                    params={"url": url}
+                )
+            elif service_type not in (enumerations.AUTO, enumerations.OWS):
+                if service_handler.service_type != service_type:
+                    raise ValidationError(
+                        _("Found service of type %(found_type)s instead "
+                          "of %(service_type)s"),
+                        params={
+                            "found_type": service_handler.service_type,
+                            "service_type": service_type
+                        }
+                    )
+            self.cleaned_data["service_handler"] = service_handler
+            self.cleaned_data["type"] = service_handler.service_type
--- a/exchange/pki/forms.py
+++ b/exchange/pki/forms.py
@@ -23,7 +23,8 @@ from geonode.services.forms import CreateServiceForm
 from geonode.services import enumerations
 from geonode.services.serviceprocessors import get_service_handler
 from django.core.exceptions import ValidationError
-from .models import SslConfig
+from .models import SslConfig, HostnamePortSslConfig
+from .utils import pki_route
 
 
 def get_ssl_config_choices():
@@ -52,6 +53,11 @@ class CreatePKIServiceForm(CreateServiceForm):
         # In order to inject the pki rerouting
         url = self.cleaned_data.get("url")
         service_type = self.cleaned_data.get("type")
+        ssl_config = self.cleaned_data("ssl_config")
+
+        HostnamePortSslConfig.objects.create_hostnameportsslconfig(url, ssl_config[0])
+        url = pki_route(url)
+
         if url is not None and service_type is not None:
             try:
                 service_handler = get_service_handler(
--- a/exchange/pki/forms.py
+++ b/exchange/pki/forms.py
@@ -55,7 +55,9 @@ class CreatePKIServiceForm(CreateServiceForm):
         service_type = self.cleaned_data.get("type")
         ssl_config = self.cleaned_data("ssl_config")
 
-        HostnamePortSslConfig.objects.create_hostnameportsslconfig(url, ssl_config[0])
+        HostnamePortSslConfig.objects.create_hostnameportsslconfig(
+            url, ssl_config[0]
+        )
         url = pki_route(url)
 
         if url is not None and service_type is not None:
--- a/exchange/pki/forms.py
+++ b/exchange/pki/forms.py
@@ -66,20 +66,20 @@ class CreatePKIServiceForm(CreateServiceForm):
                     base_url=url, service_type=service_type)
             except Exception:
                 raise ValidationError(
-                    _("Could not connect to the service at %(url)s"),
+                    "Could not connect to the service at %(url)s",
                     params={"url": url}
                 )
             if not service_handler.has_resources():
                 raise ValidationError(
-                    _("Could not find importable resources for the service "
-                      "at %(url)s"),
+                    "Could not find importable resources for the service " +
+                    "at %(url)s",
                     params={"url": url}
                 )
             elif service_type not in (enumerations.AUTO, enumerations.OWS):
                 if service_handler.service_type != service_type:
                     raise ValidationError(
-                        _("Found service of type %(found_type)s instead "
-                          "of %(service_type)s"),
+                        "Found service of type %(found_type)s instead " +
+                        "of %(service_type)s",
                         params={
                             "found_type": service_handler.service_type,
                             "service_type": service_type
--- a/exchange/pki/models.py
+++ b/exchange/pki/models.py
@@ -25,7 +25,7 @@ import logging
 from django.db import models
 from django.core.exceptions import ValidationError
 
-from exchange.pki.utils import hostname_port
+from .utils import hostname_port as filter_hostname_port
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


exchange/pki/models.py
```diff
@@ -25,6 +25,7 @@
 from django.db import models
 from django.core.exceptions import ValidationError
 
+from exchange.pki.utils import hostname_port
```
dakcarto: This should be `from .utils ...`, yes?

dakcarto: Actually, I think that's cyclical importing, since `utils` imports `models`. Maybe I need to figure out how to _not_ include `models` in `utils`.

exchange/pki/models.py
```diff
@@ -274,6 +275,22 @@ class Meta:
...
+        '''
+        Instantiates a new HostnamePortSslConfig
+        Expected to be done in creation of new service via form validation
+        :param url: The url of the service, not parsed
+        :param id: The id of the ssl config
+        :return: new HostnamePortSslConfig
+        '''
+        service_hostnameportsslconfig = self.create(
+            hostname_port=hostname_port(url),
+            ssl_config_id=id
```
dakcarto: Does setting just the `pk` work here? The `HostnamePortSslConfig` expects an SslConfig record instance now (not just an integer), i.e. [in `tests.py` I had to do this](https://github.com/boundlessgeo/exchange/blob/pki/exchange/pki/tests/tests.py#L124-L133).

exchange/pki/models.py
```diff
@@ -274,6 +275,22 @@ class Meta:
...
+class HostnamePortSslConfigManager(models.Manager):
+    def create_hostnameportsslconfig(self, url, id):
+        '''
+        Instantiates a new HostnamePortSslConfig
+        Expected to be done in creation of new service via form validation
+        :param url: The url of the service, not parsed
+        :param id: The id of the ssl config
+        :return: new HostnamePortSslConfig
+        '''
+        service_hostnameportsslconfig = self.create(
```
dakcarto: What about updates, i.e. when a record for the hostname:port already exists and just the related `SslConfig` record needs changed? Does self.create auto-check and perform an update, or do we have to check first, then decide whether it is a create or update?

Also, it the record exists, but nothing has changed, i.e. the `SslConfig` is the same, we need to not attempt a create/update. This is important, because a `HostnamePortSslConfig`, I think, should not be removed upon remote service removal: first, because another service might be dependent upon it, too (something we can check for), or we want to just leave it, so when a new service is registered with the same hostname:port the right `SslConfig` is auto-selected.

Maybe we auto-delete the record if only the service to be deleted is the _only_ service referencing it (would have to check all service URL's for whether they are `pki`-proxied). Conversely, we could just _never_ delete `HostnamePortSslConfig` records and let the admin clean them up later, in which case we would probably want to add an index to that table.

exchange/pki/utils.py
```diff
@@ -278,3 +280,16 @@ def get_ssl_context_opts(url):
...
+def pki_route(url):
+    '''
+    Reroutes a service url through our internal proxy
+    :param url: Original service url
+    :return: Modified url via internal proxy
+    Ex: url = https://myserver.com:port/geoserver/wms
+    return https://<site>/pki/myserver.com%3Aport/geoserver/wms
+    '''
+    site_url = Site.objects.get_current().domain
+    pki_url = 'https://' + site_url + '/pki' + quote(url)
```
dakcarto: Is `quote` idempotent here? I mean, is the remote service form also doing this? Maybe we just change the ':' of hostname:port to and underscore and no `quote`ing, letting remote service form handle that.

exchange/pki/utils.py
```diff
@@ -278,3 +280,16 @@ def get_ssl_context_opts(url):
...
+    return pki_url.lower()
```
dakcarto: Only the hostname:port needs to be lowercased. The rest of the URL should remain unchanged, just for passthru clarity.

exchange/pki/forms.py
```diff
@@ -0,0 +1,89 @@
...
+        url = self.cleaned_data.get("url")
+        service_type = self.cleaned_data.get("type")
+        ssl_config = self.cleaned_data("ssl_config")
+
+        HostnamePortSslConfig.objects.create_hostnameportsslconfig(
+            url, ssl_config[0]
+        )
+        url = pki_route(url)
+
+        if url is not None and service_type is not None:
```
dakcarto: By "Unfortunately we have to override ..." does that mean we can't call `super(...` for the rest of this? If `url` was a property, i.e. `self.url` that would work, but that may require a commit upstream to factor that in.

exchange/pki/forms.py
```diff
@@ -0,0 +1,89 @@
...
+
+class CreatePKIServiceForm(CreateServiceForm):
+    # In addition to normal service registration,
+    # we want to be able to associate it with an SslConfig model
+    ssl_config = forms.ChoiceField(
+        label="SSL Config",
+        choices=get_ssl_config_choices(),
+        initial='',
+    )
+
```
dakcarto: We should do a lookup on the `HostnamePortSslConfigManager` model for hostname:port, and if one is found, auto-select the appropriate SslConfig `pk`.

exchange/pki/forms.py
travislbrundage: At first glance no, but there may be a workaround, I'm going to revisit and see if there is a way.

exchange/pki/utils.py
travislbrundage: Unfortunately I don't think it is since there is an `unquote()`, however, I don't think there should be any problem with doing it? In other words if I did `quote(quote(url))` it has the same result as just `quote(url)`, and I don't believe the remote service form is doing it currently which is why I put it here. I could make the assumption it's already done or just look for `:` instead though

exchange/pki/utils.py
travislbrundage: Okay, will adjust that.

exchange/pki/models.py
travislbrundage: I'm not entirely sure, cyclical importing gets a bit confusing. I believe it should work as long as there's not some circular dependency in the functions.

exchange/pki/utils.py
dakcarto: Ok, we can experiment with the MapProxy endpoint and see how it interacts and what the URL looks like in the logs. Probably a non-issue to `quote()`.

exchange/pki/forms.py
travislbrundage: After investigating, I don't believe there is a sane workaround. The problem is that while the url is passed in to a service handler, upon creation of the appropriate service in the `get_service_handler` function, it will do things with that url in its initialization function (in this case, `WmsServiceHandler` and `MapserverServiceHandler` both use the url in different ways to affect the final creation of the model via `__init__`). So while I could hack around this, it gets messy and at best is slightly confusing, but I think it's probably not worth it and this approach is more straightforward, although unfortunate.

If the service handler models mentioned above were designed to play more nicely with updating the url it would be easy, but they're not.

exchange/pki/models.py
dakcarto: Note, in the above referenced _test_ code...
```
ssl_config = SslConfig.objects.get(pk=pk)
```
is unprotected by a `try` conditional on purpose. It should not be in this context, but that's obvious, I suppose.

exchange/pki/models.py
travislbrundage: Ah I think I see th eproblem here. Multiple things are being called `hostname_port`. I'll adjust the import.

exchange/pki/models.py
dakcarto: Oh, `ssl_config_id` should be `ssl_config`. Maybe you were testing this against older code. I changed the field name when moving from a `pk` integer field to a ForeignKey field.

dakcarto: Also, a `try` conditional will mostly protect against the rare case where an admin has actually _removed_ the SSL config _in between_ a user loading the registration page (with a widget that still has a `pk` to the now deleted config) and when the user finally submits the registration.

exchange/pki/models.py
travislbrundage: Right, I thought I should just do a `get_or_create`, but that actually won't work either. I think I need to try to get the model first if it exists, then assign the new `SslConfig`. If it doesn't exist, create it.

Unsure about deletion, but I think that shouldn't affect this part?

exchange/pki/models.py
travislbrundage: Yeah, I think I'll need to do that - didn't notice the change, thanks for pointing it out

exchange/pki/models.py
dakcarto: Yeah, I don't think deletion factors in either.

exchange/pki/models.py
travislbrundage: It didn't fix the problem - I thought python was a bit smarter about its importing than it is. It seems to be complaining about the cyclical import (although it's just giving me vague `ImportError` with no explanation, I'm fairly certain it's because of this).

exchange/pki/models.py
```diff
@@ -274,6 +275,31 @@ class Meta:
...
+class HostnamePortSslConfigManager(models.Manager):
+    def create_hostnameportsslconfig(self, url, pk):
+        '''
+        Instantiates a new HostnamePortSslConfig
+        Expected to be done in creation of new service via form validation
+        :param url: The url of the service, not parsed
+        :param pk: The pk id of the ssl config
+        :return: new HostnamePortSslConfig
+        '''
+        ssl_config = SslConfig.objects.get(pk=pk)
```
dakcarto: This needs a try conditional, or will fail with DoesNotExist.

exchange/pki/utils.py
```diff
@@ -278,3 +280,16 @@ def get_ssl_context_opts(url):
...
+def pki_route(url):
+    '''
+    Reroutes a service url through our internal proxy
+    :param url: Original service url
+    :return: Modified url via internal proxy
+    Ex: url = https://myserver.com:port/geoserver/wms
+    return https://<site>/pki/myserver.com%3Aport/geoserver/wms
+    '''
+    site_url = Site.objects.get_current().domain
+    pki_url = 'https://' + site_url + '/pki' + quote(url).lower()
```
dakcarto: Instead of `.lower()` on all of `url`. Pass it through `pki.utils.normalize_hostname(url)`, which lowercases just the hostname component, leaving the query, etc. unaffected.

dakcarto: Prior to `quote(url)`, of course. 😉

exchange/pki/models.py
```diff
@@ -284,7 +285,12 @@ def create_hostnameportsslconfig(self, url, pk):
         :param pk: The pk id of the ssl config
         :return: new HostnamePortSslConfig
         '''
-        ssl_config = SslConfig.objects.get(pk=pk)
+        # If the user tried to give a bad SslConfig,
+        # just assign it to the default
+        try:
+            ssl_config = SslConfig.objects.get(pk=pk)
+        except SslConfig.DoesNotExist:
+            ssl_config = SSL_DEFAULT_CONFIG
```
dakcarto: Oh. `SSL_DEFAULT_CONFIG` is [just a dict](https://github.com/boundlessgeo/exchange/blob/pki/exchange/pki/settings.py), not an `SslConfig` instance. We would need to create an `SslConfig` record, based upon `SSL_DEFAULT_CONFIG` and then use that. However, I'm not sure that is the right approach.

Review the following to see how `SSL_DEFAULT_CONFIG` is used now: https://github.com/boundlessgeo/exchange/blob/pki/exchange/pki/ssl_adapter.py#L178-L204

I think we _do not_ want a `HostnamePortSslConfig` created with a default SSL config, if this `try` block fails. Maybe the error should just be allowed to raise, then we log it and then let the registration fail (informing the user somehow)?

Side note: maybe we just change `SSL_DEFAULT_CONFIG` to _actually be_ an `SslConfig` instance, and add it to the `models` module, or leave it in `settings` (though I don't think `settings` should import `models`).

travislbrundage: Hmmm, right that's a good point. So if this creation fails, it means somehow it was attempted to select a `pk` of an `SslConfig` that does not exist - so raising an error makes sense. I'm not sure how it reports to the user, though.

This should be a very hard thing to make happen in the first place, so I think as long as an error is raised it's fine.

dakcarto: Ok, thinking about this some more... This may fix a couple issues.

If we do have a `SSL_DEFAULT_CONFIG` that is actually an SslConfig instance, we could prepend that to the SslConfig `.all()` output and load that into the selector, maybe as 'Secure Default'. Then, there will _always_ be a default secure SSL config, ready to insert into the SslConfig model table, that a user can choose, *regardless of whether an admin has added any SslConfigs*. In other words, a default this is better than basic `requests` package SSL.

Again, this should still be optional for all `https://` registrations, where the user still has to _opt in_ by selecting even the default one. Probably the default for the widget should be `----------` or something similar, e.g. `---select----` (whatever the norm is there).

Then, later, if the SsLConfig associated with a hostname:port has for some reason been removed, the `SSL_DEFAULT_CONFIG` instance can be auto-saved to the SslModel with a name like 'Secure Default copy for <hostname:port>'. Basically, like your code above, but with auto-creation based upon the default.

Then, if an admin updates the system with new SslConfigs, they, or the user, can just 'realign' the existing registered service to a some new config. The user would do it on the registered service details page, while the admin can do it more efficiently via the admin console.

dakcarto: Adding the default to the selector widget, we should probably set it to `-1` for its `pk`, so the code above knows what to do if the user is not selecting an existing SslConfig, i.e. that a new default record needs auto-created.

